PR #11, opened 2019-07-04
Support custom datadog tags

srvaroa:
This change expands the Datadog spec allowing users to inject tags that
will get added to the DD_TAGS environment variable in the DD sidecar.

For example:

    metrics:
      datadog:
        enabled: true
          tags:
            tag1: value1
            tag2: value2

Will inject "tag1:value1,tag2:value2" in the DD_TAGS environment
variable in the Datadog sidecar.

Signed-off-by: Galo Navarro <anglorvaroa@gmail.com>


tests/fiaas_deploy_daemon/deployer/kubernetes/deployment/test_datadog.py
```diff
@@ -43,7 +43,8 @@ def test_noop_when_not_enabled(self, datadog, app_spec, deployment):
 
     @pytest.mark.parametrize("best_effort_required", (False, True))
     def test_adds_env_when_enabled(self, datadog, app_spec, deployment, best_effort_required):
-        app_spec = app_spec._replace(datadog=True)
+        datadog_spec = app_spec.datadog._replace(enabled=True, tags=dict())
```
srvaroa: Should this be tags = None?

tests/fiaas_deploy_daemon/deployer/kubernetes/test_ingress_deploy.py
```diff
@@ -29,7 +30,7 @@ def app_spec(**kwargs):
         admin_access=False,
         secrets_in_environment=False,
         prometheus=PrometheusSpec(enabled=True, port='http', path='/internal-backstage/prometheus'),
-        datadog=False,
+        datadog=DatadogSpec(enabled=False,tags=dict()),
```
srvaroa: Ditto, should this be tags=None?

tests/fiaas_deploy_daemon/deployer/kubernetes/deployment/test_datadog.py
xavileon: I don't think. Empty dict should be fine as default, as in https://github.com/fiaas/fiaas-deploy-daemon/pull/11/files#diff-13d0cf28b75b00450ea41240032087a4R31 you're assuming that tags is a dict (with None that line would fail).

docs/v3_spec.md
```diff
@@ -423,6 +423,7 @@ Default value:
 metrics:
   datadog:
     enabled: false
+    tags:
```
birgirst: This should be 
`tags: {}`

fiaas_deploy_daemon/specs/v3/defaults.yml
```diff
@@ -56,6 +56,7 @@ metrics:
     path: /_/metrics # Path to prometheus-metrics
   datadog:
     enabled: false # if true, attach a datadog sidecar
+    tags: # A list of custom tags to add to all metrics exposed by the pods
```
birgirst: tags: {} # A map of custom tags to add to all metrics exposed by the pods

tests/fiaas_deploy_daemon/deployer/kubernetes/deployment/test_datadog.py
srvaroa: Yeah, I was wondering if we should actually support None at datadog.py (e.g. for backwards compatibility reasons with v2).  Thanks!

fiaas_deploy_daemon/specs/v3/defaults.yml
srvaroa: Thanks!, fixed

tests/fiaas_deploy_daemon/specs/v3/data/examples/full.yml
```diff
@@ -38,6 +38,9 @@ metrics:
     path: /prometheus-metrics-here
   datadog:
     enabled: true
+    tags:
+      tag1: value1
+      tag2: value2
```
oyvindio: This test data is used by both `test_v3factory.py` and `test_e2e.py`. The e2e test essentially runs a deployment using the input file (this file) into a minikube instance, and asserts that the resulting resources match manifests generally stored here: https://github.com/fiaas/fiaas-deploy-daemon/tree/master/tests/fiaas_deploy_daemon/e2e_expected. See https://github.com/fiaas/fiaas-deploy-daemon/blob/master/tests/fiaas_deploy_daemon/test_e2e.py#L150-L154 for the test definition using this file. I think the environment variables of the resulting deployment manifest will need to be updated for the test to eventually pass.

As an aside, it looks like semaphore either doesn't build branches from forks or is missing the required configuration to do so, otherwise the CI build status should show up on the PR

tests/fiaas_deploy_daemon/deployer/kubernetes/deployment/test_datadog.py
```diff
@@ -57,7 +58,9 @@ def test_adds_env_when_enabled(self, datadog, app_spec, deployment, best_effort_
         ("rincewind", "discworld")
     ))
     def test_adds_container_when_enabled(self, datadog, app_spec, deployment, best_effort_required, name, namespace):
-        app_spec = app_spec._replace(datadog=True, name=name, namespace=namespace)
+        datadog_spec = app_spec.datadog._replace(enabled=True, tags=dict())
```
oyvindio: I think it would be nice to have a unit test that covers the code path that handles tags added in the fiaas config too

oyvindio:
Since semaphore didn't build this branch since it is on a fork, I tried pushing the commits as a regular branch, and now it looks like the CI build status showed up on the PR too, probably because the commit refs are the same

srvaroa:
@oyvindio I get a 404 at the details link from the check above (https://fiaas-svc.semaphoreci.com/workflows/ecb05eea-8aaa-48d8-8e6f-03ca49e5a92b), maybe permissions?

tests/fiaas_deploy_daemon/specs/v3/data/examples/full.yml
srvaroa: Thanks! I did not run the e2e tests locally so I could not spot this. I'll get that added.

oyvindio:
> @oyvindio I get a 404 at the details link from the check above (https://fiaas-svc.semaphoreci.com/workflows/ecb05eea-8aaa-48d8-8e6f-03ca49e5a92b), maybe permissions?

I have added you to the github org and to semaphore now, so I think you should have access to the CI logs. You should also be able to push to a branch on this repo now, which is a bit more convenient since semaphore doesn't seem to build forks

srvaroa:
Repushing as a banch in this repo so CI can vote.

PR #18, opened 2019-07-29
Support custom datadog tags

srvaroa:
This change expands the Datadog spec allowing users to inject tags that will get added to the DD_TAGS environment variable in the DD sidecar.

For example:

    metrics:
      datadog:
        enabled: true
        tags:
          tag1: value1
          tag2: value2

Will inject "tag1:value1,tag2:value2" in the DD_TAGS environment variable in the Datadog sidecar.  Note that this includes a small change to make the order of DD_TAGS predictable, in order to ease tests.


srvaroa:
Comes from https://github.com/fiaas/fiaas-deploy-daemon/pull/11

tests/fiaas_deploy_daemon/deployer/kubernetes/deployment/test_datadog.py
```diff
@@ -57,15 +58,24 @@ def test_adds_env_when_enabled(self, datadog, app_spec, deployment, best_effort_
         ("rincewind", "discworld")
     ))
     def test_adds_container_when_enabled(self, datadog, app_spec, deployment, best_effort_required, name, namespace):
-        app_spec = app_spec._replace(datadog=True, name=name, namespace=namespace)
+        datadog_spec = app_spec.datadog._replace(
```
srvaroa: @oyvindio from #11 you had this comment:

> I think it would be nice to have a unit test that covers the code path that handles tags added in the fiaas config too

I was looking into https://github.com/fiaas/fiaas-deploy-daemon/blob/master/fiaas_deploy_daemon/config.py and my reading is that there is no option to make a namespace-wide config for anything Datadog other than the DATADOG_CONTAINER_IMAGE.  E.g. I cannot mandate "datadog.enabled = true" which already existed.  If that's the case, since there is no support yet for datadog.* properties would the test you refer to still be relevant?

I may be very well missing some context here, so pointers would be appreciated.

Thanks!

oyvindio: I just meant that it would be useful that there was a test at the unit level that ran the code path that uses the custom tags and tests that it comes out the way it is supposed to in the resulting API call. The current version of this test does exactly that as far as I can see, so that is fine 👍

srvaroa:
@birgirst @mortenlj I moved #11 here, it's pretty much the same change with all review comments fixed. Let me know if it looks mergeable, thanks!

srvaroa:
@oyvindio we've verified this on our cluster so from our part it's good to promote to stable.

PR #24, opened 2019-09-03
Only create one TLS entry per ingress/application

oyvindio:
When using `extensions.tls`, FIAAS  currently creates one TLS entry per hostname in addition to one TLS entry which contains all hostnames. This results in requesting `# hostnames + 1` certificates when using an automated certificate provisioning tool like kube-lego or cert-manager. 
This change disables creation of the per hostname entries by default, and hides that behavior behind a configuration flag (`--enable-deprecated-tls-entry-per-host`). That way we only keep the single TLS entry per ingress/application, which reduces the number of neccessary certificates drastically. I added the configuration flag in an attempt to make this change easier to roll out gradually.


tests/fiaas_deploy_daemon/test_e2e.py
```diff
@@ -177,6 +177,12 @@ def ready():
                 Ingress: "e2e_expected/tls-ingress-cert-issuer.yml",
                 HorizontalPodAutoscaler: "e2e_expected/tls-hpa-cert-issuer.yml",
             }),
+            ("v3/data/examples/tls_enabled_multiple.yml", {
+                Service: "e2e_expected/tls-service-multiple.yml",
+                Deployment: "e2e_expected/tls-deployment-multiple.yml",
```
mortenlj: This file seems to be missing.

oyvindio: It is not missing, but I have apparently missed deleting the service,deployment and hpa entries in this test case, fixing it now

oyvindio:
I wasn't 100% sure about this, so I tested this in minikube and it seems to work as expected. Adding the config flag to a version of fiaas-deploy-daemon without this patch works fine too, as long as it is done via `cluster_config.yaml` or environment variable; CLI flag will not work.

mortenlj:
Should this be merged?

xavileon:
@mortenlj talking with @oyvindio a while back, we decided for us (Adevinta) to handle the merge of this one as this might create some issues with currently deployed applications in our cluster. Let us know if you need this merged for some reason so we can discuss how we handle that on our side.

mortenlj:
@xavileon : I was just wondering, as leaving PRs open for a long time might be a sign of something forgotten etc. It's fine to leave it until you're ready.

PR #26, opened 2019-09-17
Update the list of kubernetes versions to run e2e tests on

mortenlj:
The new list of tested kubernetes versions is v1.9.11, v1.10.12, v1.12.10 and v1.14.4.

While updating the list, I also had to change how we do e2e tests, as I couldn't get minikube to work across all versions we wanted. I have instead switched to using the kubernetes-in-docker setup from bsycorp/kind.

Since we no longer test versions older than 1.9, we are also no longer testing a version that supports Third Party Resources. I have therefore removed the support for TPRs.


birgirst:
👏 

There are still some references to `paasbetaapplications` and 1.6 in `docs/operator_guide.md` that we could probably remove as well.

Comment in `fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py` mentioning 1.6 and 1.7.

`bin/README.md` references 1.6 and `bin/dind-cluster-v1.6.sh` may perhaps be removed?

oyvindio:
I'm having trouble getting the e2e tests to run on macOS with docker for mac. The tests all seem to fail when trying to start fiaas-deploy-daemon. It seems to have something to do with mounting the test working directory into the fiaas-deploy-daemon container failing. I'm not sure if this is a general problem or an issue with my setup, but I think my docker install is fairly standard.

```――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― ERROR at setup of TestE2E.test_custom_resource_definition_deploy[v1.9.11-ClusterIP-data/v2minimal.yml] ――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
[gw1] darwin -- Python 2.7.15 /Users/oyvoverg/src/github/fiaas-deploy-daemon/.tox/py27/bin/python

self = <test_e2e.TestE2E object at 0x1088e72d0>, request = <SubRequest 'fdd' for <Function 'test_custom_resource_definition_deploy[v1.9.11-ClusterIP-data/v2minimal.yml]'>>
kubernetes = {'api-cert': '/var/folders/mw/gzkbv7x921j40y3lk_ly5yf8ljzy9g/T/kind-ClusterIP_v1.9.11_8fb82807-1751-4471-b82a-ed72175e.../T/kind-ClusterIP_v1.9.11_8fb82807-1751-4471-b82a-ed72175eceaa-jfGiQl/client_key', 'server': 'https://localhost:32777'}, service_type = 'ClusterIP', k8s_version = 'v1.9.11'
use_docker_for_e2e = <function dockerize at 0x108f23398>

    @pytest.fixture(scope="module")
    def fdd(self, request, kubernetes, service_type, k8s_version, use_docker_for_e2e):
        port = get_unbound_port()
        args = [
            "fiaas-deploy-daemon",
            "--port", str(port),
            "--api-server", kubernetes["server"],
            "--api-cert", kubernetes["api-cert"],
            "--client-cert", kubernetes["client-cert"],
            "--client-key", kubernetes["client-key"],
            "--service-type", service_type,
            "--ingress-suffix", "svc.test.example.com",
            "--environment", "test",
            "--datadog-container-image", "DATADOG_IMAGE",
            "--strongbox-init-container-image", "STRONGBOX_IMAGE",
            "--use-ingress-tls", "default_off",
        ]
        if crd_supported(k8s_version):
            args.append("--enable-crd-support")
        cert_path = os.path.dirname(kubernetes["api-cert"])
        args = use_docker_for_e2e(request, cert_path, service_type, k8s_version, port) + args
        fdd = subprocess.Popen(args, stdout=sys.stderr, env=merge_dicts(os.environ, {"NAMESPACE": "default"}))
        time.sleep(1)
        if fdd.poll() is not None:
>           pytest.fail("fiaas-deploy-daemon has crashed after startup, inspect logs")
E           Failed: fiaas-deploy-daemon has crashed after startup, inspect logs

tests/fiaas_deploy_daemon/test_e2e.py:91: Failed
------------------------------------------------------------------------------------------------------------------------------------------------------------ Captured stderr setup ------------------------------------------------------------------------------------------------------------------------------------------------------------
docker: Error response from daemon: invalid mount config for type "bind": bind source path does not exist: /var/folders/mw/gzkbv7x921j40y3lk_ly5yf8ljzy9g/T/kind-ClusterIP_v1.9.11_8fb82807-1751-4471-b82a-ed72175eceaa-jfGiQl.
See 'docker run --help'.```

mortenlj:
I'll try to figure out what could be causing that, but it might be a docker for mac thing, in which case it will be hard for me to debug.

mortenlj:
@oyvindio : From what I can gather reading https://docs.docker.com/docker-for-mac/osxfs/, mounts should work, if the source is in one of these locations: `/Users/`, `/Volumes/`, `/private/`, and `/tmp` (by default, you can add others). It seems pythons `mkdtemp` creates a temporary directory under `/var/` instead of in `/tmp/`, causing it to not be exported to the docker for mac VM.. I'm not sure how to handle this...? Is it normal for macOS to have a `/tmp`, but not use it for temp data? (Is Python doing the wrong thing, or Docker-for-mac?)

oyvindio:
I see, that explains why it would fail. On macOS, it seems that `$TMPDIR` is set to a directory that is only read/write-able by `$USER`, I assume as a security measure. This is `/var/folders/...` above. There is also a `/tmp`, which is writable by anyone. I guess `mkdtemp` in python will use `$TMPDIR`, which results in a path that is not available inside the docker for mac VM. 

[This documentation](https://developer.apple.com/library/archive/documentation/Security/Conceptual/SecureCodingGuide/Articles/RaceConditions.html) seems to indicate that you can use `/tmp` if you really have to, although it seems to be discouraged for the same reasons that `$TMPDIR` doesn't point there. For this particular purpose, maybe a solution could be to always use `/tmp` as the prefix for mkdtemp on macOS?

It is also possible to add directories to expose to the Docker for mac VM, but that probably requires adding all of `/var` or `/var/folders`. Also it would be something that anyone trying to run these tests on macOS would need to do before the tests can pass.

PR #27, opened 2019-09-23
Add license headers

mortenlj:
This commit simply adds a standard Apache 2.0 license header to all source files where it makes sense (let me know if I missed any). This is the result of a script, so some weirdness can occur, point them out if you see things that are wrong.

### DIFF
--- a/.semaphore/latest.yml
+++ b/.semaphore/latest.yml
@@ -1,18 +1,17 @@
 
 # Copyright 2017-2019 The FIAAS Authors
-# 
+#
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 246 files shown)


mortenlj:
Crap. It's because of these changes. Codestyle issues, because the script used added trailing whitespace to some of the lines. Weirdly enough, it did not complain about this locally.. I'll fix.

mortenlj:
To fix it, I had to ask IntelliJ to reformat all the code, which ended up doing lots of other code formatting changes as well (optimize imports, clean up blank lines here and there etc). No actual code changes.

mortenlj:
The latest failures are because of minikube, so I'm going to merge this now.

PR #31, opened 2019-09-26
Make Kind e2e tests run on Docker for mac

oyvindio:
~Builds on #26~

- Always use /tmp for temporary test data when running e2e tests on macOS, as discussed in #26.
- Due to network differences on Docker for mac vs. native Docker on Linux, some more changes were necessary to make the tests run: 

We are running Kind in a container, and need to be able to connect to its apiserver from fiaas-deploy-daemon either running in a container or on the host, and from the tests themselves, which always runs on the host. Kind exposes the apiserver on port 8443 in the container it runs, and we also map this to `localhost:$RANDOM_PORT` on the host that runs the container.

On Linux, we can run Docker with host networking. This runs all containers on the hosts network, meaning we can connect to ports bound on localhost whether we're running inside a container (fiaas-deploy-daemon) or not (the tests themselves), and it will work fine. On macOS with Docker for mac, it is not possible to use host networking, so we have to use different endpoints when connecting to the apiserver running in one container from the host and from fiaas-deploy-daemon running in another container. This commit changes the kubernetes['server'] endpoint into kubernetes['container-to-container-server'] for connecting from fiaas-deploy-daemon running in a container, and kuberentes['host-to-container-server'] for connecting from fiaas-deploy-daemon running on the host, and for connecting from the tests themselves.

The TLS certificate Kind uses is pre-generated, and has `localhost` and `kubernetes` among others as valid SANs. To make sure we can validate the cert when connecting from fiaas-deploy-daemon in a container, we use `--add-host` to map `kubernetes` to the Kind container's IP in the fiaas-deploy-daemon container's hosts file.

### DIFF
--- a/tests/fiaas_deploy_daemon/utils.py
+++ b/tests/fiaas_deploy_daemon/utils.py
@@ -247,7 +247,9 @@ class KindWrapper(object):
     def __init__(self, k8s_version, name):
         self.k8s_version = k8s_version
         self.name = name
-        self._workdir = tempfile.mkdtemp(prefix="kind-{}-".format(name))
+        # on Docker for mac, directories under $TMPDIR can't be mounted by default. use /tmp, which works
+        tmp_dir = '/tmp' if _is_macos() else None
+        self._workdir = tempfile.mkdtemp(prefix="kind-{}-".format(name), dir=tmp_dir)
         self._client = docker.from_env()
         self._container = None
 
@@ -305,3 +307,6 @@ class KindWrapper(object):
         with open(path, "wb") as fobj:
             fobj.write(raw_data)
         return path
+
+def _is_macos():
+    return os.uname()[0] == 'Darwin'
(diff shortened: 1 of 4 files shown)


tests/fiaas_deploy_daemon/utils.py
```diff
@@ -263,10 +265,17 @@ def start(self):
             client_cert = self._save_to_file("client_cert", config["users"][-1]["user"]["client-certificate-data"])
             client_key = self._save_to_file("client_key", config["users"][-1]["user"]["client-key-data"])
             result = {
-                "server": "https://localhost:{}".format(api_port),
+                # "server": "https://kubernetes:8443",
```
mortenlj: This looks like a left-over line.

oyvindio: yep, I'll remove it

mortenlj:
Ironic that the e2e tests failed.. The push build worked fine, and only some tests failed in the pr build, so it's a matter of unstable tests and not something wrong in the PR. I'll rerun the build.

PR #36, opened 2019-10-15
Write some documentation on using CRDs for deployment

mortenlj:
I have written a short section on how to use CRDs for deployment, since that's the primary method and we don't actually document how that works. During this I found that I needed to describe the objects, and instead of referring to Python code, or inventing my own schema description I just used JSON Schema.

CRDs have an optional field for describing the object using JSON Schema, so we should consider moving this file to somewhere we can refer to it, and attach it to the CRDs when creating them. We also should have a more structured approach to the definition and creation of our CRDs.

It might be natural to start thinking about moving some of this documentation to the github pages repository, creating a proper documentation site.

### DIFF
--- a/docs/operator_guide.md
+++ b/docs/operator_guide.md
@@ -28,7 +28,7 @@ The Basics
 
 In order for FIAAS to function in a cluster, some basics needs to be in place
 
-* Autoscaling based on CPU metrics requires Heapster, deployed in the kube-system namespace
+* Autoscaling based on CPU metrics requires the metrics API (provided by Heapster or metrics-server), deployed in the kube-system namespace
 * Applications expect DNS to work, so kube-dns, coredns or an equivalent substitute should be installed
 * Log aggregation, so that stdout and stderr from applications are collected and aggregated and collected somewhere it can be found. Typically Fluentd collecting and sending to a suitable storage.
 * An ingress controller capable of handling all the required features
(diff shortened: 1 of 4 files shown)


docs/operator_guide.md
```diff
@@ -28,7 +28,7 @@ The Basics
 
 In order for FIAAS to function in a cluster, some basics needs to be in place
 
-* Autoscaling based on CPU metrics requires Heapster, deployed in the kube-system namespace
+* Autoscaling based on CPU metrics requires the metrics API (provided by Heapster or metrics-server), deployed in the kube-system namespace
```
oyvindio: I'm not sure heapster or metrics-server *have* to be deployed in the kube-system namespace. Maybe remove that part?

docs/operator_guide.md
```diff
@@ -154,6 +154,30 @@ See the Kubernetes documentation about [emptyDir](https://kubernetes.io/docs/con
...
 
 
+Deploying an application
+------------------------
+
+To deploy an application with FIAAS, you need three things:
+
+- A name for your application (Should follow the [Kubernetes conventions for names](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names))
+- A docker image reference
+- A fiaas configuration for your application (commonly referred to as `fiaas.yml`)
```
oyvindio: nitpick: fiaas should be uppercase

PR #37, opened 2019-10-16
Fix scale up issue when autoscaler is enabled and application downscaled

adriengentil:
This fix aims to set the number of replicas to replicas.maximum when the
current deployment has 0 replica set. Otherwise, the number of replicas stays
to 0 and the application is never upscaled (the HPA is disabled with this
state: "scaling is disabled since the replica count of the target is zero").

### DIFF
--- a/fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
+++ b/fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
@@ -98,9 +98,11 @@ class DeploymentDeployer(object):
         if should_have_autoscaler(app_spec):
             try:
                 deployment = Deployment.get(app_spec.name, app_spec.namespace)
-                replicas = deployment.spec.replicas
-                LOG.info("Configured replica size (%d) for deployment is being ignored, as current running replica size"
-                         " is different (%d) for %s", app_spec.replicas, deployment.spec.replicas, app_spec.name)
+                # the autoscaler won't scale up the deployment if the current number of replicas is 0
+                if deployment.spec.replicas > 0:
+                    replicas = deployment.spec.replicas
+                    LOG.info("Configured replica size (%d) for deployment is being ignored, as current running replica size"
+                             " is different (%d) for %s", app_spec.replicas, deployment.spec.replicas, app_spec.name)
             except NotFound:
                 pass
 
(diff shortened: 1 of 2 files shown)


mortenlj:
This behavior is actually intentional. The thought being that when the deployment is scaled to 0 (or anything outside the normal range of the HPA), then the application developer has done that on purpose and it's not our job to overrule them. 

This might be one of the decisions that is worth discussing again, if you have a particular problem you want to solve. My current stance is that there is bigger potential for harm in activating a deactivated application than in not doing anything (they scaled it down manually, they can scale it up manually).

adriengentil:
Thanks for your reply!

In that case, we scaled down the application using fiaas, setting replicas.minimum=0 and replicas.maximum=0. So we were expecting that setting the values back to replicas.minimum=2 and replicas.maximum=3 (for example) would scale up the application.

It's not a usual usecase, so we can live with it.

xavileon:
Just to elaborate a little bit, the confusion for the user came because they were changing the FIAAS Application object, not the managed Deployment object. This behaviour contradicts what we always say about the FIAAS Application object being the source of truth. We are fine saying that's a special case and if they want to scale down to 0 better romove the Application object altogether, but just wanted to clarify.

mortenlj:
I hadn't thought of that scenario. Setting your min/max to 0 in fiaas.yml will effectively disable your app and require manual intervention to get it to run again (almost, you can set it to min==max==1 first, to get it running, then set min/max as you want them). In that case, this behavior is weird. And I agree with the sentiment that we should strive to make Application the source of truth.

I'd like to think more on this, because I also like the current behavior.

oyvindio:
In a strict operator model where the desired state is described by the application resource, and fiaas continually moves the actual state towards the desired state, I think the proposed change makes more sense than the current behavior. At the same time there might be some app developers who will have to adapt to this behavior since manually `kubectl scale`-ing the application to 0 replicas will no longer be persistent. That might be ok though since the current behavior is sort of confusing unless you know that you have to manually scale the application back up before deploying if you want the deployment to actually do anything.

An idea (I'm not saying this should be scoped into this PR) to help make that transition easier might be to provide tooling to modify the replica count in the application resource, maybe via small script or even a kubectl plugin (`kubectl fiaas scale` ?).

adriengentil:
Can this PR be merged?

Issue #29, opened 2019-09-24
Teams and tags are not supported when deploying using CRDs


mortenlj:
In the kafka deploy method used at FINN, we support setting teams and tags labels on the deployed objects. This is used to track metrics and ownership across deployments.

When deploying using CRDs, we have not implemented support for these labels, and instead hardcode them to empty lists.

For FINN to move to CRD deployments, this needs to be supported. For us to get to python3 (#6) we need to get rid of the kafka code, which can only happen when FINN has completed the move to CRDs.

oyvindio:
The requirement in this issue is to make the existing teams and tags mechanism in the pipeline consumer possible to configure via the Application CRD.
There is a somewhat similar requirement that is currently solved in mast: In mast there is some code that allows the orchestrating system (typically Spinnaker) to set some annotations that is injected into the `annotations` key in the FIAAS config that is used in the Application resource. Currently these are used to add traceability metadata like links to the commit and CI build that triggered the deploy.

At the risk of increasing the scope here, maybe this could be generalized to a mechanism to set some global metadata (labels and/or annotations) that is applied to all resources fiaas generates, which could solve both these requirements? This will move the responsibility of setting the teams and tags labels properly from fiaas-deploy-daemon to the orchestrating system, or the system that sits in between and creates the Application resources. This will mean that the team and tags labels will not necessarily look the same (or even exist) everywhere FIAAS is used.

mortenlj:
The Spinnaker <-> Mast thing uses annotations, but I guess we could handle annotations and labels in the same way. Would the natural thing be to add two new keys to `spec` then?

How should we model this? Do we need to be able to specify labels and annotations per object (like in `fiaas.yml`), or should we just have one set of annotations/lables that gets applied to everything?

A while back someone (I forget who) talked about wanting to be able to configure labels/annotations to apply per namespace (in `cluster_config.yml`). Would it be natural to tackle that feature while we are at it?

oyvindio:
Another use case for annotations that are propagated to all resources fiaas manages could be: Set a callback URL on the application resource which is propagated to the applicationstatus, which can  be used later by some other tool to report status for the deployment back to the initiating system.  Much like what `pipeline.Reporter` currently does, but enabling keeping the implementation outside of fiaas-deploy-daemon.

oyvindio:
[...]

PR #42, opened 2019-10-28
Consumer needs to send additional labels and annotations too

mortenlj:
When implementing #29, we forgot to test the consumer code (the part that is FINN specific, and doesn't have e2e tests).

This solves that issue by setting the parameters to `None`.


mortenlj:
I'll just merge this to get things working again. The remaining build is expected to go green.

PR #45, opened 2019-11-11
Use self-built kind for 1.9

gregjones:
bsycorp/kind don't push updated versions for 1.9 any more, and
the latest push for 1.9.11 was done a little over a year ago, with
a certificate set to expire after 1 year, hence now kubelet doesn't
start cleanly.

fiaas/kind is a rebuild of the last commit that supported 1.9: 9cc87f6
In case it need to be re-built: the image was built by running:

```
export DOCKER_IMAGE=18.06.1-ce-dind
export KUBERNETES_VERSION=v1.9.11
./build.sh fiaas/kind:latest-1.9 fiaas/kind:$KUBERNETES_VERSION
docker login # as 'captainfiaas'
docker push fiaas/kind:$KUBERNETES_VERSION
```

Note that KUBERNETES_VERSION should exist as an env var as well as the
positional arg to the command. And DOCKER_IMAGE is set because later
versions don't work (this version is the one that was 'stable-dind' at
the time v1.9.11 was built).

### DIFF
--- a/tests/fiaas_deploy_daemon/utils.py
+++ b/tests/fiaas_deploy_daemon/utils.py
@@ -243,6 +243,9 @@ def get_unbound_port():
 
 class KindWrapper(object):
     DOCKER_IMAGE = "bsycorp/kind"
+    # bsycorp/kind for 1.9 isn't being updated, and the latest version has an expired cert
+    DOCKER_IMAGE_19 = "fiaas/kind"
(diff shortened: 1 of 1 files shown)


oyvindio:
It looks like certificates in the self-built image has expired now, since it was pushed a year ago now. I have rebuilt the image again and pushed it to the same tag - e2e tests seem to work again now

PR #48, opened 2019-11-21
Refactor lifecycle signals

### LINKED ISSUE #34: Refactor the Lifecycle signals
We use blinker to handle the various lifecycle events in fiaas-deploy-daemon. These are deploy initiated, started, failed and success. Initially, we used to send the AppSpec to these signal handlers, but changed to individual parameters when we introduced the initiated signal. This is because at the time of initiated, we don't have everything we need to create an AppSpec.

As time passed we have added more data, and now the list of parameters to send on each signal is getting long and complicated. We are also sending the exact same set of parameters on each signal, and all signal handlers effectively handle the signal in the same way with simply a string parameter indicating which signal was actually received.

I suggest to refactor the signals as follows:
- Collapse all signals to a single signal, with an additional parameter `state`. The `state` parameter indicates which of the four states just happened.
- Create a `namedtuple` in the lifecycle module called `Subject` which is meant to describe the subject of the signal. This object should contain all the current parameters sent to the signals
- Create the Subject when sending the initiated signal
- Attach the Subject to AppSpec when creating it, so that subsequent signals can use the already created Subject

This will simplify quite a bit of code. It will move the logic for selecting data to send to the signals to one place, and the signal handling will be simpler.

gregjones:
The list of parameters to the signals has grown quite long, and each
received the same set of parameters, and the handlers essentially handled
each in the same way. This switches to use a single signal, and replaces
the list of parameters with a 'Subject' that wraps them, along with a status.

Fixes #34

### DIFF
--- a/tests/fiaas_deploy_daemon/crd/test_crd_watcher.py
+++ b/tests/fiaas_deploy_daemon/crd/test_crd_watcher.py
@@ -31,7 +31,7 @@ from fiaas_deploy_daemon.config import Configuration
 from fiaas_deploy_daemon.crd import CrdWatcher
 from fiaas_deploy_daemon.crd.types import FiaasApplication, AdditionalLabelsOrAnnotations
 from fiaas_deploy_daemon.deployer import DeployerEvent
-from fiaas_deploy_daemon.lifecycle import Lifecycle
+from fiaas_deploy_daemon.lifecycle import Lifecycle, Subject
 from fiaas_deploy_daemon.specs.factory import InvalidConfiguration
 
 ADD_EVENT = {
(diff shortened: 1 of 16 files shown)


tests/fiaas_deploy_daemon/crd/test_crd_watcher.py
```diff
@@ -173,6 +173,9 @@ def test_deploy(self, crd_watcher, deploy_queue, spec_factory, watcher, app_spec
 
         app_spec = app_spec._replace(name=app_name, namespace=namespace, deployment_id=deployment_id)
         spec_factory.return_value = app_spec
+        lifecycle_subject = None
```
mortenlj: Maybe remove this line?

gregjones: Ah, yeah. Left over from a previous attempt. It's gone with 84ca20f72438c15331725f39ad61039c1df68517

PR #49, opened 2019-11-21
Log errors before saving status

gregjones:
It would be useful to have these messages included in the logs
that get saved along with the status to aid debugging errors.

### DIFF
--- a/fiaas_deploy_daemon/deployer/deploy.py
+++ b/fiaas_deploy_daemon/deployer/deploy.py
@@ -74,6 +74,7 @@ class Deployer(DaemonThread):
                 self._bookkeeper.success(app_spec)
             LOG.info("Completed deployment of %r", app_spec)
         except Exception:
+            LOG.exception("Error while deploying %s: ", app_spec.name)
             self._lifecycle.failed(app_name=app_spec.name,
                                    namespace=app_spec.namespace,
                                    deployment_id=app_spec.deployment_id,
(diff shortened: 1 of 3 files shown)


fiaas_deploy_daemon/deployer/deploy.py
```diff
@@ -74,14 +74,14 @@ def _update(self, app_spec):
                 self._bookkeeper.success(app_spec)
             LOG.info("Completed deployment of %r", app_spec)
         except Exception:
+            LOG.exception("Error while deploying %s: ", app_spec.name)
```
xavileon: is the message of the exception logged? might be useful to have as well?

gregjones: `logging.exception` automatically includes the stack-trace when it's inside an `except`

xavileon: ah I see, then it's ok. Just a comment: not sure if it's doable, but do you think it makes sense to use "error" then instead with the actual exception message rather than the full stack trace so the user is presented with something not scary? Then we can leave the log.exception after the bookeeping for internal/operator debugging.

gregjones: I think changing what is logged to account for that would be a little odd, and the intention could easily be missed. So I've changed the formatter that's used for the captured logs, so it skips the trace and includes just the last line. It leaves the logs to stdout untouched.

gregjones:
@mortenlj or @oyvindio would you mind taking a look at this, in case there's a reason separate of mast to keep some stacktraces in the captured logs?

PR #52, opened 2019-11-29
Keep existing annotations and labels on ApplicationStatus when updating the resource

### LINKED ISSUE #50: Labels and annotations on ApplicationStatus resources are overwritten on update
fiaas-deploy-daemon overwrites labels and annotations on the ApplicationStatus resources it manages when it updates these. We have a use case where it would be useful to store some information in a label or annotation on ApplicationStatus resources, to avoid duplicate reporting to the system that orchestrates deployments, so it would be useful if annotations and labels set on the ApplicationStatus were persistent.

oyvindio:
Keep existing annotations and labels on ApplicationStatus when updating the resource to allow external systems to store metadata on ApplicationStatus resources which is persistent across updates made by fiaas-deploy-daemon. Fixes #50


oyvindio:
I restarted of the CI jobs because it failed due to what looks like a transient issue with starting kind for the e2e test

PR #55, opened 2019-12-13
Specify args for initiate to allow for optionals to be omitted

### LINKED ISSUE #54: TypeError while trying to deploy via pipeline consumer
The pipeline consumer raises TypeError when trying to process deployments:

```Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/fiaas_deploy_daemon/base_thread.py", line 29, in _logging_target
    self()
  File "/usr/local/lib/python2.7/site-packages/fiaas_deploy_daemon/pipeline/consumer.py", line 66, in __call__
    self._handle_message(deploy_counter, message, message_counter)
  File "/usr/local/lib/python2.7/site-packages/fiaas_deploy_daemon/pipeline/consumer.py", line 76, in _handle_message
    deployment_id=self._deployment_id(event))
  File "/usr/local/lib/python2.7/site-packages/fiaas_deploy_daemon/lifecycle.py", line 39, in initiate
    subject = Subject(*args, **kwargs)
TypeError: __new__() takes exactly 7 arguments (4 given)```

gregjones:
This fixes #54

### DIFF
--- a/fiaas_deploy_daemon/lifecycle.py
+++ b/fiaas_deploy_daemon/lifecycle.py
@@ -35,8 +35,8 @@ class Lifecycle(object):
     def change(self, status, subject):
         self.state_change_signal.send(status=status, subject=subject)
 
-    def initiate(self, *args, **kwargs):
-        subject = Subject(*args, **kwargs)
+    def initiate(self, app_name, namespace, deployment_id, repository=None, labels=None, annotations=None):
+        subject = Subject(app_name, namespace, deployment_id, repository, labels, annotations)
         self.state_change_signal.send(status=STATUS_INITIATED, subject=subject)
         return subject
 


gregjones:
Changing the tests for the components that currently mock Lifecycle to use an actual instance with a mocked signal might help avoid issues with this - I started looking at that, but it seems a bit more involved, maybe it shouldn't delay the fix.

oyvindio:
> Changing the tests for the components that currently mock Lifecycle to use an actual instance with a mocked signal might help avoid issues with this - I started looking at that, but it seems a bit more involved, maybe it shouldn't delay the fix.

Since this change, as long as Lifecycle mocks are created with `create_autospec(..., spec_set=True)`, an exception will be raised if a required i.e. non-keyword argument is missing. It wouldn't have caught this issue though, since `lifecycle.initiate` accepted any args before this commit. I'd say that is good enough; what do you think?

gregjones:
Ah, ok. I didn't realise the mock was that smart... I tried to make the test fail with a slightly different version, and then didn't retry that after this final version of the change. 👍

PR #57, opened 2019-12-18
Add script to run local fiaas-deploy-daemon against a kind Kubernetes cluster

oyvindio:
Works with the latest version of https://github.com/kubernetes-sigs/kind

- add run_fdd_against_kind
- remove remains of ThirdPartyResource support from run_fdd_against_minikube


oyvindio:
> Maybe a note in the readme about how devs should set things up so that this works?

Good point, I've added a short note about how to use this script

PR #60, opened 2020-01-07
Fix missing repository from usage-reporting

gregjones:
This fixes the issue introduced by #48 that resulted in 'repository'
being missing from the usage-reporting events.


gregjones:
This is the same as was in this removed code: https://github.com/fiaas/fiaas-deploy-daemon/pull/48/files#diff-295fd1fdb9dea6ffb8f0e3c79c2f9afeL96 (which was previously different to the `_repository` method in crd/watcher.py.

PR #64, opened 2020-01-10
Update to support kubernetes 1.16

### LINKED ISSUE #39: Transition from extensions/v1beta1 Deployment to apps/v1 Deployment
FIAAS currently uses extensions/v1beta1 Deployment. This resource is deprecated in Kubernetes 1.16, and will be removed in Kubernetes 1.18.
Code paths that handle deployments need to be updated to use apps/v1 Deployment resources to be compatible with Kubernetes 1.18 and later.

gregjones:
Updates the version of k8s to the latest, which has the changes
to support the new API URLs for Deployments. Also includes 1.16
as a version in the end-to-end tests.

Fixes #39

### DIFF
--- a/tests/fiaas_deploy_daemon/conftest.py
+++ b/tests/fiaas_deploy_daemon/conftest.py
@@ -187,6 +187,6 @@ def _open():
         yield mock_open
 
 
-@pytest.fixture(scope="session", params=("v1.9.11", "v1.10.12", "v1.12.10", "v1.14.4"))
+@pytest.fixture(scope="session", params=("v1.9.11", "v1.10.12", "v1.12.10", "v1.14.4", "v1.16.3"))
 def k8s_version(request):
     yield request.param
--- a/tests/fiaas_deploy_daemon/conftest.py
+++ b/tests/fiaas_deploy_daemon/conftest.py
@@ -187,6 +187,6 @@ def _open():
         yield mock_open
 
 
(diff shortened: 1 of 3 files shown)


tests/fiaas_deploy_daemon/conftest.py
```diff
@@ -187,6 +187,6 @@ def _mock_namespace_file_open(name, *args, **kwargs):
         yield mock_open
 
 
-@pytest.fixture(scope="session", params=("v1.9.11", "v1.10.12", "v1.12.10", "v1.14.4"))
+@pytest.fixture(scope="session", params=("v1.9.11", "v1.10.12", "v1.12.10", "v1.14.4", "v1.16.3"))
```
mortenlj: Any chance we can drop 1.9.11 support? (FINN is on 1.12, moving to 1.14 and 1.16 soon)

Every added version increases the time to perform tests, so we don't want this list to grow too big.

gregjones: We still have a 1.9 cluster 😢 
Dropping 1.10 would be fine for us, we have 1.9 and 1.14 and the next ones will be 1.16 (or maybe higher, tbd...). What do you think?

mortenlj: Let's drop 1.10 then. :+1:

gregjones: 9a0430d does that

Issue #33, opened 2019-09-29
e2e tests are hard to debug


mortenlj:
With the switch to using the kind container for e2e tests, we have a new type of instability in our e2e tests, where we get connection reset when communicating with kind. Since we don't capture any logs, that's all we know.

We should look into capturing logs from the kind container, so that we have more insights into why we get connection reset. Then maybe we can find solutions to avoid connection issues in the future, making the tests more stable.

mortenlj:
I have experienced this issue locally, while working on #68. One observation is that I don't get any logs from the container, and at one point I saw a message in the terminal that mentioned something about a non-existent container. Unfortunately, a second later, the test success output (the line with green checks or red crosses) was written on top of the line with the container error, so I didn't see what it said in detail. I've tried finding out where such a message would come from, but with no luck.

If we assume that I caught the correct meaning from this line (and that I didn't just imagine it, even though I can't explain where it would have come from), the problem might be related to the container "going away" in some way. I guess the container might die some time after startup, and then any currently connected client would probably get the connection reset. However, you would have thought there would be some logs in this case, but that doesn't seem to be the case.

Anyway, I'm going to continue investigating this theory for a little while.

mortenlj:
Fixed(ish) in #68.

PR #68, opened 2020-01-17
Dump related logs from kind container at test completion

mortenlj:
This will print logs from the kind container to stdout in two different scenarios:

If there is an exception when starting the kind container, we dump the full logs from the container. This should hopefully provide information about why kind failed to start.

In the tests, I have introduced a context manager, which notes the time it was entered, and at exit will dump the logs from the kind container that has occured in the interval between start and now. This should be all logs relevant for the current test item. Most of the time, this seems to be empty.

This should provide much needed information about what is going on inside the kind container during the tests. However, I have now experienced the "Connection reset by peer" failure mentioned in #33 locally, after these changes were present in my code, and I got no logs at all. I will write a comment on that issue with some observations.

I still think this change is useful, but it's not going to help solve #33 :disappointed:


mortenlj:
This change only applies to the e2e test by the way, I see I neglected to mention that.

PR #77, opened 2020-03-31
Fix version in example

gregjones:


gregjones:
This is mostly to see if a new merge to master will un-stick Semaphore, which seems not to like the latest events there.

README.md
```diff
@@ -96,7 +96,7 @@ spec:
   application: example
   image: nginx:1.13.0
   config:
-    version: 2
+    version: 3
     host: example.com
     prometheus:
       enabled: false
```
oyvindio: To be a valid v3 config the host and prometheus keys have to be moved under ingress and metrics respectively, something like this:
```yaml
ingress:
  - host: example.com
metrics:
  prometheus
    enabled: false
```

gregjones: Uff. Thanks - fixed it :)

PR #79, opened 2020-04-01
Change datadog imagePullPolicy for "latest"

gregjones:
This changes the imagePullPolicy for the datadog sidecar, based
on the tag of the image so that "latest" or no tag will result in
"Always". This makes it consistent with the setting applied for the
main container.

### DIFF
--- a/tests/fiaas_deploy_daemon/deployer/kubernetes/deployment/test_datadog.py
+++ b/tests/fiaas_deploy_daemon/deployer/kubernetes/deployment/test_datadog.py
@@ -24,7 +24,8 @@ from k8s.models.pod import PodTemplateSpec, PodSpec, Container, EnvVar
 from fiaas_deploy_daemon import Configuration
 from fiaas_deploy_daemon.deployer.kubernetes.deployment import DataDog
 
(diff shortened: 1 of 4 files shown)


tests/fiaas_deploy_daemon/deployer/kubernetes/deployment/test_datadog.py
```diff
@@ -107,3 +108,17 @@ def test_adds_container_when_enabled(self, datadog, app_spec, deployment, best_e
...
+
+        datadog_spec = app_spec.datadog._replace(
+            enabled=True
+        )
+        app_spec = app_spec._replace(datadog=datadog_spec)
+
+        datadog.apply(deployment, app_spec, False)
+
+        actual = deployment.as_dict()["spec"]["template"]["spec"]["containers"][-1]
+        assert actual['image'] == CONTAINER_IMAGE_LATEST
```
mortenlj, line 124: This doesn't look like it is actually testing that the image pull policy is set correctly?

gregjones: That's very true. It helps if I save the file before committing! 🙈

Issue #71, opened 2020-02-24
Generalize secret support in FIAAS


xavileon:
FIAAS currently supports importing secrets into pods in two different ways:
- native k8s secrets (mounting the secret with the same name as the FIAAS app)
- importing from strongbox via an init container

It would be helpful to generalize how the init container for strongbox works, define a very simple interface for it, so we could swap in any other secret implementation. By doing so, we could support other secret managements tools like AWS Secret Manager or Vault.

mortenlj:
Actually, there are three ways. The third one is a somewhat generic init container variant that is (was?) used by FINN to implement Vault support.

It might be worth trying to make the Strongbox and Generic init container solutions into one solution that is truly generic, but I'm not sure how easy that will be. There is a reason we didn't use the existing generic init container solution for Strongbox (I think it's because Strongbox requires additional configuration).

xavileon:
Ah cool! maybe that's all we need then 👍

gregjones:
I wasn't aware of that generic option either, it looks good, but might need more config options... 🙀 
I'm also wondering whether the flow for strongbox might actually work with secret-manager with no changes, as probably all it needs is the IAM role and the group could be re-purposed?

An issue might be for people migrating, that the image is set at the namespace-level, but maybe an image that works with both could somehow be made... 🤔

gregjones:
So we had some discussions internally about this, and we think the following might be a nice way to support the cases we'd like to, namely:

- Allow the choice of where to find secrets to be part of the app-spec, to allow for migrations away from strongbox
- Support AWS-based stores authenticated via credentials provided by kiam (which requires the pod to have a specific annotation)

We extend paas.yml so it can take the following:
```yaml
extensions:
  secrets:
    type: aws-parameter-store
    parameters: # arbitrary k-v pairs set as env on the init-container
      AWS_REGION: eu-central-7
    annotations: # annotation that will be added to the secret-init-container
      iam.amazonaws.com/role: arn::etc.
```

The 'type' would be a reference to an image stored in the fiaas-deploy-daemon config, e.g.

```yaml
cluster_config.yaml: |-
[...]
  secret-init-containers:
    aws-parameter-store: containers.example.com/foo/parameter-store-init-container:latest
    strongbox: containers.example.com/foo/strongbox-store-init-container:latest
```

[...]

PR #80, opened 2020-04-01
Refactor init-container secrets

gregjones:
This rewrites the processing of the default init-container,
and the strongbox version, to be based on the same implementation
that will be re-usable for the new functionality proposed in #71.

The initialisation of an empty list of `secret_specs` in Secrets.apply will be
replaced by the secrets specified in the app-spec, and the map of
`_available_secrets_containers` in GenericInitSecrets will be replaced
by a value taken from the deploy-daemon config.

### DIFF
--- a/fiaas_deploy_daemon/deployer/kubernetes/deployment/secrets.py
+++ b/fiaas_deploy_daemon/deployer/kubernetes/deployment/secrets.py
@@ -20,26 +20,44 @@ from k8s.models.pod import EnvFromSource, SecretEnvSource, EnvVar, Container, Co
     EmptyDirVolumeSource, ConfigMapVolumeSource, SecretVolumeSource
 
 from fiaas_deploy_daemon.tools import merge_dicts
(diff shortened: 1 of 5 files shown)


fiaas_deploy_daemon/deployer/kubernetes/deployment/secrets.py
```diff
@@ -20,26 +20,44 @@
...
+                           parameters={},
+                           annotations={})
 
     def apply(self, deployment, app_spec):
-        if self._secrets_image_set:
-            self._generic_init.apply(deployment, app_spec)
-        elif self._uses_strongbox_init_container(app_spec):
-            self._strongbox.apply(deployment, app_spec)
+        secret_specs = []
+        if not secret_specs:
```
adriengentil, line 52: I feel this check is useless here, the list was declared empty just before

adriengentil: it makes sense in the second PR.

PR #81, opened 2020-04-09
Refactor CI setup

mortenlj:
- Removed files that are no longer relevant/used
- Made docker_build script usable for local development
- Made tox integration tests use code directly


mortenlj:
As this was building I realised one more thing I'd like to fix, which is that there really is no need to push the docker image on every build on every branch. Only master branch builds needs to push the image.

PR #84, opened 2020-04-20
Send datadog global tags

gerardsegarra:
This change accepts the new parameter global-datadog-tags, which adds the given key:value to the Datadog tags for all the containers.
For example, if you want to send in which cluster or region the container is, you can deploy fiaas-deploy-daemon with this information instead of changing every service to send it to Datadog.


gerardsegarra:
Closing again and creating a new one to trigger CI

PR #91, opened 2020-05-19
Make deployments default to minimum replicas

amanya:
FIAAS sets deployment replicas to max (from the resources definition) by default. This is done with the expectation that tha HPA will reduce the number of pods to the minimum/correct number with time. However, for those environments where there's no HPA, in our case all dev envs and most pre envs, downscaling never happens. This leaves apps with an unnecessary high number of pods.


fiaas_deploy_daemon/specs/v3/factory.py
```diff
@@ -47,7 +47,7 @@ def __call__(self, name, image, teams, tags, app_config, deployment_id, namespac
             namespace=namespace,
             name=name,
             image=image,
-            replicas=lookup["replicas"]["maximum"],
+            replicas=lookup["replicas"]["minimum"],
```
xavileon, line 50: is this field used now?

amanya: Yes, i.e. [here](https://github.com/fiaas/fiaas-deploy-daemon/blob/master/fiaas_deploy_daemon/deployer/kubernetes/autoscaler.py#L45) but maybe it can be replaced by `app_spec.autoscaler.min_replicas`

tests/fiaas_deploy_daemon/deployer/kubernetes/test_autoscaler.py
```diff
@@ -31,20 +31,20 @@ def test_default_spec_should_create_no_autoscaler(app_spec):
 
 
 def test_autoscaler_enabled_and_1_replica_gives_no_autoscaler(app_spec):
-    app_spec = app_spec._replace(autoscaler=AutoscalerSpec(enabled=True, min_replicas=2, cpu_threshold_percentage=50))
+    app_spec = app_spec._replace(autoscaler=AutoscalerSpec(enabled=True, min_replicas=1, max_replicas=2, cpu_threshold_percentage=50))
```
gregjones, line 34: This makes the test confusing; it should be verifying that if min=max=1 then no autoscaler, but if they're different, then there _should_ be an autoscaler

fiaas_deploy_daemon/specs/v3/factory.py
xavileon: that [line](https://github.com/fiaas/fiaas-deploy-daemon/blob/master/fiaas_deploy_daemon/deployer/kubernetes/autoscaler.py#L45) needs to be `app_spec.autoscaler.max_replicas` indeed, because you're setting replicas to min here. If that is the only place, I'd remove it from the app spec to avoid confusion on its meaning. WDYT?

tests/fiaas_deploy_daemon/deployer/kubernetes/test_autoscaler.py
```diff
@@ -31,20 +31,20 @@ def test_default_spec_should_create_no_autoscaler(app_spec):
...
     app_spec = app_spec._replace(replicas=1)
     assert should_have_autoscaler(app_spec) is False
 
 
 def test_autoscaler_enabled_and_2_replica_and_no_requested_cpu_gives_no_autoscaler(app_spec):
-    app_spec = app_spec._replace(autoscaler=AutoscalerSpec(enabled=True, min_replicas=2, cpu_threshold_percentage=50))
+    app_spec = app_spec._replace(autoscaler=AutoscalerSpec(enabled=True, min_replicas=2, max_replicas=2, cpu_threshold_percentage=50))
```
gregjones, line 40: Similar here, setting min=max means there shouldn't be an autoscaler, so this isn't making it test the right property of the condition

tests/fiaas_deploy_daemon/deployer/kubernetes/test_autoscaler.py
```diff
@@ -31,20 +31,20 @@ def test_default_spec_should_create_no_autoscaler(app_spec):
...
     app_spec = app_spec._replace(replicas=2)
 
     assert should_have_autoscaler(app_spec) is False
 
 
 def test_autoscaler_enabled_and_2_replica_and__requested_cpu_gives_autoscaler(app_spec):
-    app_spec = app_spec._replace(autoscaler=AutoscalerSpec(enabled=True, min_replicas=2, cpu_threshold_percentage=50))
+    app_spec = app_spec._replace(autoscaler=AutoscalerSpec(enabled=True, min_replicas=2, max_replicas=2, cpu_threshold_percentage=50))
```
gregjones, line 47: Also here, if min=max, then the logic should mean there isn't an autoscaler.

fiaas_deploy_daemon/specs/v3/factory.py
amanya: Oh right, will do that :+1:

oyvindio:
It is not entirely clear to me what you mean by "environments where there's no HPA because no enforcement is done"; could you try to explain? 
My understanding of the rationale for wanting to make this change is; you have applications which are deployed in multiple environments with the same fiaas config (with # replicas scaled for production?). In some of these environments you do not have HPA, and end up with applications running max replicas. Is this accurate? If I'm misunderstanding or missing something, please let me know

amanya:
Hi @oyvindio, sorry for the late reply. Your assumption is right, I tried to clarify the description.

fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
```diff
@@ -94,23 +94,23 @@ def deploy(self, app_spec, selector, labels, besteffort_qos_is_required):
...
-                             " is different (%d) for %s", app_spec.replicas, deployment.spec.replicas, app_spec.name)
+                             " is different (%d) for %s", app_spec.autoscaler.min_replicas, deployment.spec.replicas, app_spec.name)
             except NotFound:
                 pass
 
         deployment_strategy = DeploymentStrategy(
             rollingUpdate=RollingUpdateDeployment(maxUnavailable=self._max_unavailable,
                                                   maxSurge=self._max_surge))
-        if app_spec.replicas == 1 and app_spec.singleton:
+        if app_spec.autoscaler.min_replicas == 1 and app_spec.singleton:
```
gregjones, line 113: This doesn't look right: before, if 'replicas' was 1, we knew that min_replicas would also have been 1 (max can't be less than min), but here you could have min=1, max=5, and in that case singleton should be ignored, as it doesn't make sense to set it. So I think this should be checking max_replicas. (singleton defaults to true, so this is important - it's not just to deal with the case where a user could set something that doesn't make sense)

tests/fiaas_deploy_daemon/deployer/kubernetes/test_deployment_deploy.py
```diff
@@ -391,7 +387,7 @@ def create_expected_deployment(config,
 
     max_surge = u"25%"
     max_unavailable = 0
-    if app_spec.replicas == 1 and app_spec.singleton:
+    if app_spec.autoscaler.min_replicas == 1 and app_spec.singleton:
```
gregjones, line 390: For the same reason as [here](https://github.com/fiaas/fiaas-deploy-daemon/pull/91/files#diff-0a3d5fdfbdba0e1dfd5133b968f65167L113), I think this should be max_replicas, or explicitly that min=max=1.

fiaas_deploy_daemon/deployer/kubernetes/ready_check.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, app_spec, bookkeeper, lifecycle, lifecycle_subject, config):
         self._lifecycle_subject = lifecycle_subject
         self._fail_after_seconds = _calculate_fail_time(
             config.ready_check_timeout_multiplier,
-            app_spec.replicas,
+            app_spec.autoscaler.max_replicas,
```
gregjones, line 35: If we're going to set replicas=min in the no-HPA environments, then it will never reach max, so this will never report success. I think it should be min_replicas

xavileon: this is only to calculate the amount of time to wait I think? So setting min or max will only have an impact on how long we wait for all pods to be ready. If that's the case, changing to min_replicas may fail too early in the case where HPA scales to the max replicas and the difference between min and max is significant?

fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
xavileon: I agree 👍

fiaas_deploy_daemon/deployer/kubernetes/ready_check.py
gregjones: Ah, yeah, you're right 👍

fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
amanya: Fixed in https://github.com/fiaas/fiaas-deploy-daemon/pull/91/commits/2ad9bab78d2e30a411424f1da6f54d37db418449

tests/fiaas_deploy_daemon/deployer/kubernetes/test_deployment_deploy.py
amanya: Fixed in https://github.com/fiaas/fiaas-deploy-daemon/pull/91/commits/2ad9bab78d2e30a411424f1da6f54d37db418449

oyvindio:
Now that I'm pretty sure that I understand the rationale; I don't think this is the best solution for this problem. I'm not 100% sure of the exact mechanics of the interaction beween horizontal pod autoscaler and deployment during a rolling upgrade, but I think these are separate control loops. Because of that I think that this may cause availability issues during rollout by scaling the deployment down before the HPA can scale it back up. I.e. not that the application would be completely unavailable, but that there might not be enough instances to handle the traffic. Depending on configuration, application and traffic, a delay before scale up could then cause the traffic to overwhelm the remaining instances.

I did a simple test locally with kind/Kubernetes 1.16, using a deployment with 3 replicas (rollingUpgrade with maxSurge=50% and maxUnavailable=0) and autoscaling from 1 to 3 replicas. I tested changing a label on the pod template (to trigger rolling upgrade) and setting replicas back to 1 at the same time to emulate the suggested behavior in this PR. The observed behavior in this case was; one pod in the new replicaset is created while two pods in the old replicaset is terminated (at this point we have only one ready pod, but the maxUnavailable=0 constraint is still fullfilled since we now have replicas=1), and then there is a delay of a few seconds before the HPA scales the deployment back up to 3. If this was a request driven application, the remaining one instance would have to handle the full traffic load during that period.

mortenlj:
> Because of that I think that this may cause availability issues during rollout by scaling the deployment down before the HPA can scale it back up.

I don't think that is what would happen, because when updating the deployment, whatever replicas it has set will be kept, ie. the current value as set by the HPA will be kept.

xavileon:
Agree with Morten and Albert. Setting it to the minimum only happens on the 1st time a deployment object is deployed. After that, subsequent deployments doesn't modify the number of replicas of a deployment [if the deployment object is managed by the HPA](https://github.com/fiaas/fiaas-deploy-daemon/blob/master/fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py#L103).

oyvindio:
Okay, I must have misunderstood something here, anyway I'm happy to be wrong about that. I think this change makes sense given that it only happens on deployment resource creation or if the deployment does not use autoscaling.

It looks like there is an issue with e2e test data that needs to be updated to reflect the new behavior, which should be fixed before this can be merged.

amanya:
Thanks for looking at it, Guys! I fixed the integration tests values, please review it again when you have a moment

PR #93, opened 2020-05-19
[WIP] Set ownerReferences

gregjones:
This adds a new service, OwnerReferences can be used to set the ownerReferences field on a k8s Model passed to it, using the relevant information from an AppSpec.

For the e2e tests, to allow the checking of the set uid field to that of the Application, it's modified to set that value on the expected object.


gregjones:
This needs https://github.com/fiaas/fiaas-deploy-daemon/pull/92

If you prefer to browse commit by commit:

-  76a8a8120207a95cb800aee61e80e9f7d7f8ccc0 adds the new service, but doesn't use it.
- a51cce5a99f59c5229f907f23a864246dcf4b7bc adds it to the Ingress objects, and includes the modifications to the tests. 
- The remaining commits are for each of the other object types

(I would squash these before merging.)

PR #94, opened 2020-05-20
Set uid=None when processing events in 'consumer' workflow

gregjones:
Here there is no Application CRD to take an ID from; when we come to
set the ownerReference, it will have to be omitted for this mode of
operation, and we can use uid=None to indicate that.


gregjones:
This fixes an issue introduced with #92 that would affect the daemon running as the event-consumer.

Issue #10, opened 2019-07-01
Populate ownerReferences in the children of fiaas TPR/CRD's


birgirst:
It would be great if fiaas populated the ownerReferences of resources created by fiaas. This would enable us to navigate to the fiaas definition from an ingress object for example.

https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/

mortenlj:
Related to #78

mortenlj:
@gregjones : Was this issue fixed in #95, or is there still something left?

gregjones:
I think it can be closed.

PR #95, opened 2020-05-21
Set ownerReferences

gregjones:
This adds a new service, OwnerReferences can be used to set the ownerReferences field on a k8s Model passed to it, using the relevant information from an AppSpec.

For the e2e tests, to allow the checking of the set uid field to that of the Application, it's modified to set that value on the expected object.

- 1f295df adds the new service, but without using it
- f728fa1 adds the ref to the Ingress object, and this commit includes the changes to the tests that are then used by the rest
- then there is a commit per object-type

I split it up to (hopefully) aid reviewing, as the test changes are a little noisy - I will squash before merging.

The code for handing the delete events isn't removed here, as any currently deployed applications won't have the owner set, and so wouldn't cleanup properly if they were deleted. Leaving some time between releasing this, and removing the delete flow, will allow time for those applications to be re-deployed so they set the reference. (The code is ready, I can open a separate PR so it can be looked at.)

This also doesn't yet include attaching the owner to the ApplicationStatus; those currently aren't cleaned up by the deletion event anyway, so there's no regression, and the code to support that is a little different. (It's ready, I'll open a separate PR for that too).

Part of #10

### DIFF
--- a/fiaas_deploy_daemon/deployer/kubernetes/__init__.py
+++ b/fiaas_deploy_daemon/deployer/kubernetes/__init__.py
@@ -23,6 +23,7 @@ from .autoscaler import AutoscalerDeployer
 from .deployment import DeploymentBindings
 from .ingress import IngressDeployer, IngressTls
 from .service import ServiceDeployer
+from .owner_references import OwnerReferences
 
 
 class K8sAdapterBindings(pinject.BindingSpec):
@@ -32,6 +33,7 @@ class K8sAdapterBindings(pinject.BindingSpec):
         bind("ingress_deployer", to_class=IngressDeployer)
         bind("autoscaler", to_class=AutoscalerDeployer)
         bind("ingress_tls", to_class=IngressTls)
+        bind("owner_references", to_class=OwnerReferences)
 
     def dependencies(self):
         return [DeploymentBindings()]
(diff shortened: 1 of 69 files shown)


mortenlj:
> The code for handing the delete events isn't removed here, as any currently deployed applications won't have the owner set, and so wouldn't cleanup properly if they were deleted. Leaving some time between releasing this, and removing the delete flow, will allow time for those applications to be re-deployed so they set the reference.

I'm fairly certain that when this is deployed, it will trigger a re-deploy of all currently deployed applications, so this problem will take care of itself. On the other hand, it's something to be aware of when rolling out.

oyvindio:
> I'm fairly certain that when this is deployed, it will trigger a re-deploy of all currently deployed applications, so this problem will take care of itself. On the other hand, it's something to be aware of when rolling out.

I think this guard https://github.com/fiaas/fiaas-deploy-daemon/blob/master/fiaas_deploy_daemon/crd/watcher.py#L99-L101 should avoid re-deploying everything

mortenlj:
> I think this guard https://github.com/fiaas/fiaas-deploy-daemon/blob/master/fiaas_deploy_daemon/crd/watcher.py#L99-L101 should avoid re-deploying everything

I think you are right. Forgot about that. :+1:

PR #98, opened 2020-05-27
Support deploying multiple ingresses

### LINKED ISSUE #89: Support for multiple Ingresses
Short version: We have a growing number of users who would like to have multiple Ingresses for their fiaas apps, because they want different annotations on them, that control visibility/access. Adding 'annotations' as a property to the items in the list of 'ingresses' allows us to support this, while keeping things simpler/compatible with the current behaviour by using this 'annotations' field as the distinguisher for when to group and when to separate hosts/paths into an Ingress object.

```yaml
ingress:
  - host: myservice-internal.myns-pro.cluster.com
    annotations:
      nginx.ingress.kubernetes.io/whitelist-source-range: “10.x.y.z/32”
  - host: myservice-public.myns-pro.cluster.com
    annotations:
      kubernetes.io/ingress.class: public
      nginx.ingress.kubernetes.io/whitelist-source-range: “215.a.b.c/32”
```

I've written up a longer version that explains the behaviour in a little more detail, with examples showing the impact (or lack of impact in the cases for backwards-compatibility): https://docs.google.com/document/d/12-hrIgxMEYJDw8qMeUzDar8xa_iV4Tc2CqfpVRo9WoE/edit?usp=sharing
(It's public for commenting/suggesting, let me know what emails to add if you'd like to edit)

gregjones:
This adds support for deploying multiple ingresses with FIAAS, by
separating the ingresses specified in the config by their distinct
requirements for annotations. This is useful as it means for ingress
controllers that use annotations for configuration - this can be
determining which controller is used (public/private network,
or configuration that affects behaviour (whitelisting, HTTP-auth)
per host/path.

This closes #89

### DIFF
--- a/docs/v3_spec.md
+++ b/docs/v3_spec.md
@@ -152,21 +152,22 @@ ingress:
     paths:
     - path: /
       port: http
+    annotations: {}
 ```
 
-All applications will get a set of default hosts, if the cluster operator has defined ingress suffixes. 
+All applications will get a set of default hosts, if the cluster operator has defined ingress suffixes.
 If you do not specify a host in your `ingress` configuration, these default hosts will be used.
-For example : 
+For example :
 1. `your-app.example1.com`
 2. `your-app.example2.com`
 
-When you expose a path on a host you get that one as well. For example : 
+When you expose a path on a host you get that one as well. For example :
 ```yaml
 ingress:
   - host: example.com
     paths:
     - path: /my-path
-``` 
+```
 
 If you want to customize paths for default hosts as well, you can do it as :
 ```yaml
@@ -179,8 +180,8 @@ ingress:
 
 ```
 
-This will make `/some-other-path` available on default hosts, but not on the host you provided in ingress. 
-Remember, default hosts will also contain the paths from the ingress.   
+This will make `/some-other-path` available on default hosts, but not on the host you provided in ingress.
+Remember, default hosts will also contain the paths from the ingress.
 
 ### host
 
@@ -203,7 +204,7 @@ ingress:
   - host: your-app.example.com
 ```
 
-If the operator of your cluster has configured host-rewrite rules they will be applied to the hostname given in this 
+If the operator of your cluster has configured host-rewrite rules they will be applied to the hostname given in this
 field. See [the operator guide](operator_guide.md#host-rewrite-rules) for details about how this feature works.
 
 In typical clusters, this value should be the host used by your application in production, and host-rewrite rules should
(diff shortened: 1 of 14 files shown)


gregjones:
This will need https://github.com/fiaas/k8s/pull/76

fiaas_deploy_daemon/deployer/kubernetes/ingress.py
```diff
@@ -51,38 +53,72 @@ def delete(self, app_spec):
...
         ingress_spec = IngressSpec(rules=per_host_ingress_rules + default_host_ingress_rules)
 
         ingress = Ingress.get_or_create(metadata=metadata, spec=ingress_spec)
         self._ingress_tls.apply(ingress, app_spec, self._get_hosts(app_spec))
         ingress.save()
 
+    def _delete_unused(self, labels):
+        filter_labels = {
+            "app": Equality(labels["app"]),
+            "fiaas/deployment_id": Inequality(labels["fiaas/deployment_id"])
```
xavileon, line 112: I think this need another filter: "fiaas/deployment_id": Exists()

Otherwise, it will match all ingresses from that does not have a fiaas/deployment_id (e.g. custom ingresses for that app created off-band). This feature should make these extra ingresses not necessary, but still seems like better fall on the safe side?

gregjones: Ah, true. But passing a map with duplicate keys isn't possible :/

gregjones: Maybe like this? Not sure if it's the tidiest API: https://github.com/fiaas/k8s/pull/80

xavileon: I think that solution is good. I thought about a map where the values may be lists, but I think the list of tuples better represents what the k8s API is.

fiaas_deploy_daemon/deployer/kubernetes/ingress.py
```diff
@@ -43,48 +45,89 @@ def deploy(self, app_spec, labels):
...
-            u"fiaas/expose": u"true" if _has_explicitly_set_host(app_spec) else u"false"
+        LOG.info("Creating/updating ingresses for %s", app_spec.name)
+        custom_labels = merge_dicts(app_spec.labels.ingress, labels)
+
+        # Group app_spec.ingresses to separate those with annotations
+        AnnotatedIngress = namedtuple("AnnotatedIngress", ["name", "ingress_items", "annotations"])
+        unannotated_ingress = AnnotatedIngress(name=app_spec.name, ingress_items=[], annotations={})
+        ingresses_by_annotations = [unannotated_ingress]
+        for ingress_item in app_spec.ingresses:
+            LOG.info(ingress_item)
```
amanya, line 66: It was your intention to only log the item here?

gregjones: Ah, no. Left over from debugging: removed in 98bd43771927468e6b43b4cfaafc38216690d708

gregjones:
I added some docs too for this, PTAL

docs/v3_spec.md
```diff
@@ -234,6 +235,30 @@ your application. Requests to `your-app.example.com/metrics` will go to the port
...
+ingress:
+  - host: your-app.example.com
+    paths:
+      - path: /foo
+  - host: other.example.com
+    paths:
+      - path: /foo
+    annotations:
+      some/annotation: bar
+```
```
oyvindio, line 261: The design document has some nice examples of how this feature works; do you think it makes sense to add some more of these examples here too?

PR #101, opened 2020-05-28
Remove code-path for responding to Delete events

gregjones:
Now that we are setting ownerReferences on the created objects
kubernetes will take care of deleting them for us when the application
is deleted, so we can remove all the code related to that.

The delete methods for service, ingress and autoscaler are still needed
for the case where the config changes such that they are no longer
required, but now they can be private/internal methods.


gregjones:
@oyvindio Yes, I missed that part. It's deleted with 5b79aa54f978d4ef650ae0db1c59e8e376edf453

gregjones:
Thanks for the reviews. As mentioned elsewhere, I'll wait on merging this until we are ready to trigger re-deploys of the apps in our clusters, I think it's relatively harmless being there and doing nothing in the short-term.

oyvindio:
Closed via #262

PR #103, opened 2020-06-29
Support deploying multiple ingresses

### LINKED ISSUE #89: Support for multiple Ingresses

gregjones:
This adds support for deploying multiple ingresses with FIAAS, by
separating the ingresses specified in the config by their distinct
requirements for annotations. This is useful as it means for ingress
controllers that use annotations for configuration - this can be
determining which controller is used (public/private network,
or configuration that affects behaviour (whitelisting, HTTP-auth)
per host/path.

This closes #89

### DIFF
--- a/docs/v3_spec.md
+++ b/docs/v3_spec.md
@@ -143,7 +143,8 @@ replicas:
 
 This object configures path-based/layer 7 routing of requests to the application.
 It is a list of hosts which can each have a list of path and port combinations. Requests to the combination of host and
-path will be routed to the port specified on the path element.
+path will be routed to the port specified on the path element. Setting annotations on an entry means a separate Ingress
+will be created, and will have metadata.annotations set accordingly.
 
 Default value:
 ```yaml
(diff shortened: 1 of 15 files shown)


gregjones:
The other pull-request wasn't showing the latest state of the branch, I thought closing it and reopening it might help, but now it won't let me reopen it ... I pushed during the github downtime this morning, I guess that caused a problem :/

@oyvindio I addressed your comments (I was away last week, hence the delay) - splitting out the grouping and the rename made a lot of sense so I did that 👍 (164ba6f51950bd7f64b42dfbdd71998c0939b79f) And I expanded the explanation a little in the docs, and added an extra example (2ca1de4d198304404c9204838eea8b4f26736e8e)

(@xavileon @adriengentil @amanya If you want to take another look too that would be wonderful)

gregjones:
Thanks @adriengentil - @xavileon , @oyvindio do you have any more comments?

Issue #70, opened 2020-01-30
Clean up environment variables passed to apps


mortenlj:
When fiaas deploys an application, a number of environment variables are set, from a number of sources. Some of these have been "marked" as deprecated (by virtue of a comment in the source code) for ages, some are somewhat weird FINN legacy stuff and some are useful but badly named.

I'd like to do a clean up, but it probably requires a detailed plan for how to go about it, so that users have time to update their applications. 

In addition to the ones FDD sets directly, we have the environment variables that comes from global config. Because we at some point thought it would be best to set a prefix to all environment variables supplied by fiaas, we added copies of all of them with the `FIAAS_` prefix with the intention to remove the non-prefixed ones later. We never followed through, and I think that's a good thing.

I'd like to clean this up, following these principles:
- Environment variables from global config are kept as is. It's up to the cluster operator to ensure that they are unique enough.
- Environment variables set by FDD for internal reasons, should have the `FIAAS_` prefix.
- Environment variables that are a legacy from FINN should be removed or replaced where needed (see table below)

This leads to me wanting to make these changes (this might also solve #12, since we remove the propagated variables completely):

| Name | Value | What to do with it |
|------|-------|--------------------|
| ARTIFACT_NAME | Name of application | Rename to FIAAS_APPLICATION_NAME. (for FINN, this requires some work, since ARTIFACT_NAME is use "everywhere" in common libraries) |
| CONSTRETTO_TAGS | Legacy FINN support | Remove (for FINN: move to global config) |
| FINN_ENV | Same as FIAAS_ENVIRONMENT, legacy from FINN. Considered deprecated for ages. | Remove |
| IMAGE | Name of deployed image. Could be useful for applications. | Rename to FIAAS_IMAGE |
| LOG_FORMAT | Logging format (plain/json). Propagated from FDDs own config. Legacy from FINN. | Remove (for FINN: move to global config) |
| LOG_STDOUT | Logging destination (true->stdout/false->file). Hardcoded to true. Legacy from FINN. | Remove (for FINN: move to global config) |
| VERSION | Version of deployed image (ie. tag of container). Might be useful, but we already have cases of name collision. | Rename to FIAAS_IMAGE_VERSION |
| | | |
| FIAAS_ENVIRONMENT | Propagated from FDDs config. Should describe environment. | Remove. If cluster operators needs it, they can put it in global config. |
[...]

PR #106, opened 2020-07-10
Allow overriding most environment variables via global-env

oyvindio:
I'm not entirely sure I'm happy with how this is tested, I may look into that a bit more. Here is a draft PR anyway;

As a first step for deprecating some FIAAS-managed environment variables
(see #70) , allow overriding the following environment variables via the
global-env mechanism:

- CONSTRETTO_TAGS
- FINN_ENV
- LOG_FORMAT
- LOG_STDOUT
- FIAAS_ENVIRONMENT
- FIAAS_INFRASTRUCTURE

It is still not possible to override the following, as they are
generated based on each application's configuration:

- ARTIFACT_NAME
- VERSION
- IMAGE

### DIFF
--- a/fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
+++ b/fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
@@ -41,8 +41,8 @@ class DeploymentDeployer(object):
         self._prometheus = prometheus
         self._secrets = deployment_secrets
         self._owner_references = owner_references
-        self._fiaas_env = _build_fiaas_env(config)
-        self._global_env = config.global_env
+        self._legacy_fiaas_env = _build_fiaas_env(config)
+        self._global_env = _build_global_env(config.global_env)
         self._lifecycle = None
         self._grace_period = self.MINIMUM_GRACE_PERIOD
         self._use_in_memory_emptydirs = config.use_in_memory_emptydirs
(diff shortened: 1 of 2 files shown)


fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
```diff
@@ -236,3 +234,12 @@ def _build_fiaas_env(config):
             "CONSTRETTO_TAGS": ",".join(("kubernetes-{}".format(config.environment), "kubernetes", config.environment)),
         })
     return env
+
+
+def _build_global_env(global_env):
+    """
+    global_env key/value are added as is and with the key prefix FIAAS_
```
mortenlj, line 241: This part was IMHO a misguided attempt at something I don't even remember anymore. Especially the bit about having both with and without the prefix. Either it should only be with `FIAAS_` prefix, or don't mess with them at all.
And again, since it's been like this a while, it might need a deprecation period and proper handling of the migration, so that users are not caught out.

fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
```diff
@@ -151,21 +151,17 @@ def _make_volume_mounts(self, app_spec):
...
-
-        # For backward compatibility. https://github.schibsted.io/finn/fiaas-deploy-daemon/pull/34
-        global_env = []
-        for name, value in self._global_env.iteritems():
-            if "FIAAS_{}".format(name) not in constants and name not in constants:
-                global_env.extend([EnvVar(name=name, value=value), EnvVar(name="FIAAS_{}".format(name), value=value)])
-            else:
-                LOG.warn("Reserved environment-variable: {} declared as global. Ignoring and continuing".format(name))
-        env.extend(global_env)
+        fiaas_managed_env = {
```
mortenlj, line 154: Contrary to the global envs, which is managed and named by the cluster operator, these variables are fully managed and named by fiaas, and should probably have had the `FIAAS_` prefix from the beginning. That would have avoided the issue where `VERSION` conflicts with something in Nodejs (wasn't it?).
Of course, since it's been like this a while, it might need a deprecation period and proper handling of the migration, so that users are not caught out.

oyvindio: Yes, I agree that these variables should have the prefix `FIAAS_`, and like you point out; changing that will require adding copies with the prefix, and a period of having both the current and the prefixed variable keys. Maybe it makes sense to make that change in a separate PR though?

fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
oyvindio: Yes, I think it makes sense to remove this behavior; it is mostly just confusing. If the cluster operator wants environment variables prefixed with `FIAAS_` they can just use that as the variable key when adding the variable. Like below I think I'd prefer handing that in a followup PR to this though

fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
mortenlj: Yeah, probably best to do those as separate PRs. 👍

oyvindio:
> I'm not entirely sure I'm happy with how this is tested, I may look into that a bit more. Here is a draft PR anyway;

I made an attempt at adding another test for various permutations of environment variables, but  it ended up duplicating so much of the existing code that tests it now that I think I'd rather just keep it as is.

oyvindio:
@gregjones @mortenlj do you want to take another look?

Issue #72, opened 2020-02-24
Support separate service accounts per application


xavileon:
FIAAS only supports using the default service account in Deployment objects to map the k8s API token inside the pod (if instructed to do so).

AWS EKS use service accounts to inject IAM Roles and provide fine-grained authorization to the AWS API, a similar functionality that `kiam` provides (using service accounts instead of annotations on pods).

To support this use case, FIAAS should allow binding different service accounts per application. This could work as follows:
- FDD will have a new setting (separateServiceAccountPerApplication) that will create a service account with the application name (if it doesn't exist), following the FIAAS naming convention.
- Deployment objects will bind this new service account in the Deployment object instead of the default account.

If the cluster operator wants to provide IAM roles for a given application, the operator may configure the app specific service account accordingly following https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html

Thoughts?

oyvindio:
I think this sounds like a very useful feature. Only creating the service accounts and not roles with rules makes it very flexible too in terms of building your own authorization regime.

I also think that this can be used to improve the current RBAC setup that skipper creates. Currently, skipper creates a [fiaas-controller](https://github.com/fiaas/skipper/blob/master/helm/fiaas-skipper/templates/rbac.yaml#L24) role and binds it to [`system:serviceaccounts`](https://github.com/fiaas/skipper/blob/master/helm/fiaas-skipper/templates/rbac.yaml#L77) via a clusterrolebinding. This is setup is a fair bit more permissive than necessary. If fiaas-deploy-daemon created serviceaccounts for each application, I think this could be limited down to only bind the role to the fiaas-deploy-daemon (and skipper and possibly mast) serviceaccounts in each namespace.

sa-mao:
I also this would make a very nice feature specially with AWS IRSA.

stribb:
How close are we here?

Thanks :)

oyvindio:
This feature was added in #159

PR #107, opened 2020-07-14
72 Allow separate service account per application

sa-mao:
[WIP]

### DIFF
--- a/fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
+++ b/fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
@@ -81,7 +81,7 @@ class DeploymentDeployer(object):
 
         automount_service_account_token = app_spec.admin_access
         init_containers = []
-        service_account_name = "default"
+        service_account_name = "default" if ("service_account_name" not in app_spec) else app_spec.service_account_name
 
         pod_spec = PodSpec(containers=containers,
                            initContainers=init_containers,
--- a/tests/fiaas_deploy_daemon/e2e_expected/custom_service_account-deployment.yml
+++ b/tests/fiaas_deploy_daemon/e2e_expected/custom_service_account-deployment.yml
@@ -1,0 +1,187 @@
+
+# Copyright 2017-2019 The FIAAS Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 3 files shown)


sa-mao:
Besides from allowing arbitrary service accounts per app there is actually no benefit. Do you think that would be less useful?

mortenlj:
There is the balance between flexibility and complexity to think about. If fiaas were to provide every feature of kubernetes in fiaas.yaml, it has no purpose other than complicate things. So for every new field we introduce, it is important to think about why we are adding this field, and do we really need it. 

And while this change won't be the one that suddenly makes fiaas.yaml too complex, that can be said of any addition, until one day we find that it is too complex.

For this use case, I think the mentioned discussion found that using a service account named the same as the app should work fine for most uses (and importantly, all the ones we knew needed to be solved), and adding an option nobody has asked for is probably not worth it.

sa-mao:
I would argue that in this case, allowing custom service accounts to be created outside of FIAAS and then mounted by introducing a new field is far better that creating service accounts by FDD, then trying to update them manually or by any other mean to inject whatever we need later on (I'm thinking here of the annotation with AWS IAM role Arn).

There is also the reusability argument (although I acknowledge it might be controversial in this case).

The way I see it: this is a perfect balance between flexibility and complexity.

xavileon:
> I would argue that in this case, allowing custom service accounts to be created outside of FIAAS and then mounted by introducing a new field is far better that creating service accounts by FDD

That doesn't mean it is not possible with the approach described in #72 I think. You could create a service account outside of FDD following the naming convention (same as the app name) and FDD could pick this up as is (the same way FDD does with externally created secrets). That probably imposes some restrictions on the flow when creating an app, but nothing that having a custom name doesn't impose anyway. It's more about convention over configuration than anything else.

Is there any use case in which the name of the service account actually matters?

tg90nor:
> That doesn't mean it is not possible with the approach described in #72 I think. You could create a service account outside of FDD following the naming convention (same as the app name) and FDD could pick this up as is (the same way FDD does with externally created secrets). That probably imposes some restrictions on the flow when creating an app, but nothing that having a custom name doesn't impose anyway. It's more about convention over configuration than anything else.

What about when FDD does not find a service account with the same name as the app? Should it fall back to `default`, or create a new SA?

xavileon:
> > That doesn't mean it is not possible with the approach described in #72 I think. You could create a service account outside of FDD following the naming convention (same as the app name) and FDD could pick this up as is (the same way FDD does with externally created secrets). That probably imposes some restrictions on the flow when creating an app, but nothing that having a custom name doesn't impose anyway. It's more about convention over configuration than anything else.
> 
> What about when FDD does not find a service account with the same name as the app? Should it fall back to `default`, or create a new SA?

I'd say it should create a new SA if FDD is configured to have a per-app service account. This would allow creating the SA before the app or modify it afterwards, depending on the different needs and workflows?

sa-mao:
In that case, I agree.

oyvindio:
This feature was merged in #159

PR #108, opened 2020-07-20
Add flag to disable deprecated managed env vars

xavileon:
FDD historically injected some environment variables that
describe the version, the docker image and the artifact name.
These variables however conflict with variables used by
some docker images creating unexpected behaviours.

This change adds the prefix `FIAAS_` to these managed
environment variables alongside the previous (now) deprecated
variables so they can cohexist for a while.

This change also introduces a flag to disable the deprecated
non-prefixed variables so this change can be rolled out
in a controller way.

Partially fixes: https://github.com/fiaas/fiaas-deploy-daemon/issues/70

Signed-off-by: Xavi León <xavi.leon@adevinta.com>

### DIFF
--- a/tests/fiaas_deploy_daemon/deployer/kubernetes/test_deployment_deploy.py
+++ b/tests/fiaas_deploy_daemon/deployer/kubernetes/test_deployment_deploy.py
@@ -172,6 +172,18 @@ class TestDeploymentDeployer(object):
         return mock.create_autospec(Secrets(config, None, None), spec_set=True, instance=True)
 
     @pytest.mark.usefixtures("get")
+    def test_managed_environment_variables(self, post, config, app_spec, datadog, prometheus, secrets, owner_references):
+        deployer = DeploymentDeployer(config, datadog, prometheus, secrets, owner_references)
+        env = deployer._make_env(app_spec)
+        env_keys = [var.name for var in env]
+        assert 'FIAAS_ARTIFACT_NAME' in env_keys
+        assert 'FIAAS_IMAGE' in env_keys
+        assert 'FIAAS_VERSION' in env_keys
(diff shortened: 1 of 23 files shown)


tests/fiaas_deploy_daemon/deployer/kubernetes/test_deployment_deploy.py
```diff
@@ -452,17 +453,27 @@ def create_expected_deployment(config,
...
         'LOG_STDOUT': 'true',
-        'VERSION': version,
-        'FIAAS_INFRASTRUCTURE': infrastructure,
+        'FIAAS_INFRASTRUCTURE': config.infrastructure,
         'LOG_FORMAT': 'json',
-        'IMAGE': 'finntech/testimage:' + version,
         'CONSTRETTO_TAGS': _create_constretto_tag(environment),
     }
 
+    _managed_env_variables = {
```
gregjones, line 464: I am not in love with this being such a repetition of the logic being tested 😞 
Maybe some tests of the _make_env directly could help keep the ordering/priority clear?

Or maybe a separate test that simply shows that setting 'VERSION' in the app's env will work when the flag is set (i.e. will end up being different to FIAAS_VERSION) would be good.

xavileon: yep, I agree on better not rely on the same logic in the tests. You mean something like https://github.com/fiaas/fiaas-deploy-daemon/pull/108/commits/54b3842a2c86ba31f34aa25241f696b03ca0f420?

xavileon: ping @gregjones 🙏

PR #109, opened 2020-08-13
[WIP] Opt out of default hosts

amanya:


mortenlj:
When you asked me about this the other day, I didn't quite get what you were asking. Now that I do, I think that the recently added "annotated ingresses" feature can be (ab)used to do what you want.

If I'm not mistaken, ingresses with annotations are not paired up with the default ingress suffixes. If you mark each ingress with an annotation, the default hosts are not generated. (You should probably test to see if I'm right).

It might still be better to have an explicit option for it, but if this is just a temporary need while migrating, maybe it's better to not add it to the application config?

amanya:
You are right, if you define an ingress with an annotation the default hosts are not created, thanks Morten!

PR #110, opened 2020-08-27
Update hpa scaleTargetRef apiVersion

cdemonchy:
Following pr #64, updates the scalerefTarget to use apps/v1 instead of
extensions/v1beta to locate the deployment.

It's needed to run fiaas on 1.16 cluster

### DIFF
--- a/fiaas_deploy_daemon/deployer/kubernetes/autoscaler.py
+++ b/fiaas_deploy_daemon/deployer/kubernetes/autoscaler.py
@@ -40,7 +40,7 @@ class AutoscalerDeployer(object):
             custom_labels = merge_dicts(app_spec.labels.horizontal_pod_autoscaler, labels)
             metadata = ObjectMeta(name=app_spec.name, namespace=app_spec.namespace, labels=custom_labels,
                                   annotations=app_spec.annotations.horizontal_pod_autoscaler)
-            scale_target_ref = CrossVersionObjectReference(kind=u"Deployment", name=app_spec.name, apiVersion="extensions/v1beta1")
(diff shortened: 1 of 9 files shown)


mortenlj:
I assume creation of HPA in a 1.16 cluster fails without this change? Our e2e tests do test Kubernetes 1.16.2, but apparently didn't catch this. Any thoughts on what we can do with the e2e tests to catch this kind of problem in the future?

cdemonchy:
The hpa with the old api version in scaleTargetRef is accepted by the api server.

But it doesn't work, when doing a desribe on it we see this message : 

> Events:
>   Type     Reason          Age                    From                       Message
>   ----     ------          ----                   ----                       -------
>   Warning  FailedGetScale  3m30s (x751 over 20h)  horizontal-pod-autoscaler  no matches for kind "Deployment" in group "extensions"

gregjones (approved):
👍

mortenlj:
> But it doesn't work, when doing a desribe on it we see this message :

Ouch. I don't see how to reasonably test for this kind of thing in the integration tests either.. good catch though!

mortenlj (approved):

oyvindio:
> The hpa with the old api version in scaleTargetRef is accepted by the api server.
> 
> But it doesn't work,

I'm not sure it is worth trying to make a e2e test that verifies that behavior, it would be a significant increase in moving async parts to try to get in line in the test and I think it would be difficult to make it a testcase that wouldn't be flaky.

oyvindio (approved):
This looks good, thanks for contributing!

Issue #112, opened 2020-08-28
Proposal: Multiple config-map support for fiaas


skogie:
We are using config-maps in fiaas today but it can be tedious at times updating and keeping track of all our config-maps when making changes. For example for our database config we use a different port in dev and production, and it also requires som difference in the ssl part of the connection-string. I'm wondering if this could have been solved using only a single config-map named i.e database-config and applied this config-map to all our apps via fiaas.yml, something like this: 
`---
version: 3
admin_access: true
replicas:
  maximum: 1
  minimum: 1
resources:
  requests:
    memory: 128Mi
ports:
  - target_port: 5000
healthchecks:
  liveness:
    http:
      path: /healthz
metrics:
  prometheus:
    path: /internal-backstage/prometheus
config_maps:
    - team-db-config
    - team-kafka-config`

The default can stay the same that a config-map with the same name as an application would be applied by default. And a default for the config-maps in fiaas.yml could be something like:
`config_maps: []`

As far as I'm aware multiple config-maps can be applied to a pod like this:

`spec:
  containers:
    - name: example-pod
      envFrom:
      - configMapRef:
          name: config-1
      - configMapRef:
          name: config-2`

Let me know what you think, I would love this feature 👍

skogie:
Opened PR #113 as a WIP

mortenlj:
The configmap we currently make available is made available in two ways: As environment variables, and mounted on disk at `/var/run/config/fiaas/`.
If we are adding this feature, I feel it natural that any additional configmaps are also made available in both ways.

There are a couple open questions that probably need to be discussed, and also documented properly:

1. Configmaps may have the same keys, and when exposing as environment variables, ordering matters. Where should the "default" configmap be in the order? First or last? Which value is kept when multiple configmaps has the same key?
2. Where should the configmaps be mounted on disk? They probably should be under `/var/run/config/`. Should it just use the name of the configmap (would that always work?), or does the config need to be a mapping of configmap to path?

PR #113, opened 2020-08-28
proposal for multiple configmaps #112

skogie:
WIP for issue #112 
860 tests passed with `tox -e test` 
Wasnt able to run the integration-test module it kept crashing so if there is more to be done here let me know.


fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
```diff
@@ -244,6 +245,15 @@ def _build_fiaas_env(config):
...
 
 
+def _add_config_maps(app_spec):
+    config_maps = [EnvFromSource(configMapRef=ConfigMapEnvSource(name=app_spec.name, optional=True))]
+    if app_spec.config_maps:
+        for config_map in app_spec.config_maps:
+            config_maps.append(EnvFromSource(configMapRef=ConfigMapEnvSource(name=config_map, optional=True)))
+        return config_maps
+    else:
+        return config_maps
```
tg90nor, line 255: You can put the return outside the if, and drop the else clause

tg90nor (commented):
Thanks for the contribution! Overall, this looks pretty good.

There should be some docs about this feature in https://github.com/fiaas/fiaas-deploy-daemon/blob/master/docs/v3_spec.md

I would also like to see some tests verifying the behaviour with 0, 1 and multiple configmaps defined for an app

Finally a small nit: I would like to see all occurences of `config_maps` renamed into `configmaps`, as that is how it is usually written when all lowercase

skogie:
The integration tests seem flaky on my machine if anyone could try to run them and give me feedback I would appreciate it!

tg90nor:
```
Results (893.49s):
     140 passed
________________________________________________________________________________________________________________________ summary _________________________________________________________________________________________________________________________  integration_test: commands succeeded
  congratulations :)
```

docs/v3_spec.md
```diff
@@ -680,6 +680,20 @@ Kubernetes apiserver. If in doubt, leave this disabled.
...
 ```
 
+## config_map
+
+| **Type** | **Required** |
+|----------|--------------|
+| list     | no           |
+
+Add extra configMaps from fiaas to your application, by default a configMap named the same as your application will be 
+added to your app by default.
```
gregjones, line 690: This reads a little awkwardly, maybe:
```suggestion
Add extra configMaps to your application. By default a configMap named the same as your application will be 
added.
```

fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
```diff
@@ -244,6 +245,14 @@ def _build_fiaas_env(config):
     return env
 
 
+def _add_config_maps(app_spec):
+    config_maps = [EnvFromSource(configMapRef=ConfigMapEnvSource(name=app_spec.name, optional=True))]
+    if app_spec.config_maps:
+        for config_map in app_spec.config_maps:
+            config_maps.append(EnvFromSource(configMapRef=ConfigMapEnvSource(name=config_map, optional=True)))
+        return config_maps
```
gregjones, line 253: This return is unnecessary, the one below will be enough

fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
```diff
@@ -244,6 +245,14 @@ def _build_fiaas_env(config):
     return env
 
 
+def _add_config_maps(app_spec):
+    config_maps = [EnvFromSource(configMapRef=ConfigMapEnvSource(name=app_spec.name, optional=True))]
```
gregjones, line 249: Does this mean that the same env vars defined in one of the extra configmaps will overwrite those from this one, as they're added afterwards? To me, it seems like the configmap named for the app should be the most specific. In either case, I think it should be documented.

gregjones (changes requested):
I like the idea, and it mostly looks good with just a couple of things I think should be clarified.

I also think Morten's comment from the issue, about also exposing them via volume-mounts makes sense - is there a reason not to do that?

fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
skogie: I'm not sure, I have to do some research to see what happens when it overwrites. If anyone have a good way to test this I'm all ears.

skogie:
> I like the idea, and it mostly looks good with just a couple of things I think should be clarified.
> 
> I also think Morten's comment from the issue, about also exposing them via volume-mounts makes sense - is there a reason not to do that?

I was not aware of the volume-mounts, but yeah i should add those aswell!

fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
skogie: @mortenlj Do you know how that works? Also, what happens today if you create a configMap with env-vars already defined within the cluster?

mortenlj: Found this in the documentation for the `envFrom` field:

> When a key exists in multiple sources, the value associated with the last source will take precedence. Values defined by an Env with a duplicate key will take precedence.

fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
```diff
@@ -146,6 +151,10 @@ def _make_volumes(self, app_spec):
 
     def _make_volume_mounts(self, app_spec):
         volume_mounts = []
+        if app_spec.config_maps:
```
skogie, line 154: @mortenlj @gregjones I just committed this, but I dont think this works out of the box as Morten mentioned in the issue. I read it is possible to use `subPath` to create a file for the configMap but it will probably(?) break something.  Right now each env variable is mounted to its own file like this 
```
/run/config/fiaas $ ls
ENV_CONFIG1 ENV_CONFIG2 ENV_CONFIG3
```
If I understand it correctly with subPath each configMap would be its own file, maybe it would be possible to make it mount under `/var/run/fiaas/config/<app_name>`(default configMap) and `/var/run/fiaas/config/test1` if you specify a configMap in fiaas.yml. Again, probably some breaking changes..

skogie: Another possiblity could be to leave the default configMap as is, and add the new ones under `/run/config/fiaas/<user_defined_configmap>`

skogie: something like this:
```
if app_spec.config_maps:
            for config_map in app_spec.config_maps:
                volume_mounts.insert(0, VolumeMount(name="{}-config".format(config_map),
                                                    readOnly=True, mountPath="/var/run/config/fiaas/{}/".format(config_map)))
```

fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
```diff
@@ -146,6 +151,10 @@ def _make_volumes(self, app_spec):
...
+            for config_map in app_spec.config_maps:
+                volume_mounts.insert(0, VolumeMount(name="{}-config".format(config_map),
```
mortenlj, line 156:
```
$ python -m timeit -s "volumes = []" "volumes.append('a short string')"
5000000 loops, best of 5: 46.1 nsec per loop
$ python -m timeit -s "volumes = []" "volumes.insert(0, 'a short string')"
50000 loops, best of 5: 12.4 usec per loop
```

Inserting at the start is massively more expensive than appending at the end. Since the user has total control over the ordering of these in the spec, I say just append, and document how that translates to precedence when deployed.

fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
```diff
@@ -146,6 +151,10 @@ def _make_volumes(self, app_spec):
...
+                                                    readOnly=True, mountPath="/var/run/config/fiaas/"))
```
mortenlj, line 157: Configmaps are mounted as directories, where each key in the configmap becomes a file in the directory. They probably need distinct mount paths. Assuming the configmap name is a valid directory name (I think it would be), I suggest just mounting at `/var/run/config/<configmap-name>`.

fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
mortenlj: See my comment further down.

skogie:
@mortenlj pushed some changes now, what do you think?

mortenlj:
Starting to look good. The only thing I'd like to change is expanding the documentation with detailing of how precedence works for the added configmaps as we have discussed.

mortenlj:
@skogie : Any chance you could sort out the documentation and maybe this can be merged?
@gregjones : What do you think, good?

docs/v3_spec.md
```diff
@@ -680,6 +680,20 @@ Kubernetes apiserver. If in doubt, leave this disabled.
 admin_access: False
 ```
 
+## config_map
+
+| **Type** | **Required** |
+|----------|--------------|
+| list     | no           |
+
+Add extra configMaps to your application. By default a configMap named the same as your application will be
```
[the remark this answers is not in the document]
gregjones: Also include here the details on how the configmaps turn into filepaths?

gregjones (commented):
+1 on the docs, and an extra thing I think should be included. We should also make sure CI runs, right?

skogie:
I will try to get the remaining changes out tomorrow. Missing some tests and documentation.

mortenlj:
What's the state of this PR, any movement?

skogie:
@mortenlj I have been working on getting the tests to work locally but rn i have a round-trip of around 20 minutes per integration_test-run and its still not passing here, I think everything should be correct but could you try to run it on your end?

mortenlj:
> I have been working on getting the tests to work locally but rn i have a round-trip of around 20 minutes per integration_test-run and its still not passing here, I think everything should be correct but could you try to run it on your end?

@skogie : The last run seems to fail because of codestyle issues. Should be enough to fix those. To just run the codestyle tests, use `tox -e codestyle` (which is significantly quicker than waiting for the full set of tests).

skogie:
@mortenlj I also have failing tests with `tox -e integration_test`, not sure if the CI server is running those?

skogie:
```
Results (16.41s):00:33
     866 passed​00:33
___________________________________ summary ____________________________________00:33
  codestyle: commands succeeded00:33
  test: commands succeeded00:33
  congratulations :)
```

tests seems to be running but the `tox -e integration_test` does not seem to be a part of the testing

mortenlj:
> tests seems to be running but the `tox -e integration_test` does not seem to be a part of the testing

Yes, we've hit https://github.com/fiaas/fiaas-deploy-daemon/issues/131

@oyvindio : Did you do any more research on that issue?

skogie:
@mortenlj is it possible for you to run it locally to see if they actually pass, or if there is anything else i need to fix? I don't think they are working correctly on my machine

skogie:
Fixed most of the tests now, but I still don't understand why these 2 are failing if someone could help me out: https://gist.github.com/skogie/4a786ff9ceec17a26466a3221819dbab and https://gist.github.com/skogie/0b480f01f46fd438e033134ef7a86de5

oyvindio:
> Yes, we've hit #131
> 
> @oyvindio : Did you do any more research on that issue?

It is on my list, but haven't been able to look into it further yet unfortunately. The only "workaround" I'm aware of to get the e2e tests to run before that issue is fixed, is to push the branch from the fork to a new branch on the source repository, which admittedly isn't a great solution

tests/fiaas_deploy_daemon/e2e_expected/v3full-deployment.yml
```diff
@@ -229,6 +241,16 @@ spec:
...
+      - configMap:
+          defaultMode: 420
+          name: test1
+          optional: true
+        name: test1-config
+      - configMap:
+          defaultMode: 420
+          name: test2
+          optional: true
+      - name: test2-config
```
mortenlj, line 253: A `- ` too much on this line I think. Should solve the remaining two tests I guess.

skogie:
Would be nice to try to run it on the CI-server now if possible.

mortenlj:
I triggered a build directly in Semaphore after changing some settings which I hope doesn't expose our docker secrets to all PRs... 🙂 
Also, I think perhaps, maybe, the above comment didn't work then, but should work now...

tests/fiaas_deploy_daemon/e2e_expected/v3-multiple-configmap-deployment.yml
```diff
@@ -0,0 +1,215 @@
+
+# Copyright 2017-2019 The FIAAS Authors
```
gregjones, line 2:
```suggestion
# Copyright 2017-2020 The FIAAS Authors
```

tests/fiaas_deploy_daemon/e2e_expected/v3-no-extra-configmap-deployment.yml
```diff
@@ -0,0 +1,193 @@
+
+# Copyright 2017-2019 The FIAAS Authors
```
gregjones, line 2:
```suggestion
# Copyright 2017-2020 The FIAAS Authors
```

tests/fiaas_deploy_daemon/e2e_expected/v3-single-extra-configmap-deployment.yml
```diff
@@ -0,0 +1,204 @@
+
+# Copyright 2017-2019 The FIAAS Authors
```
gregjones, line 2:
```suggestion
# Copyright 2017-2020 The FIAAS Authors
```

tests/fiaas_deploy_daemon/specs/v3/data/examples/multiple_config_maps.yml
```diff
@@ -0,0 +1,19 @@
+
+# Copyright 2017-2019 The FIAAS Authors
```
gregjones, line 2:
```suggestion
# Copyright 2017-2020 The FIAAS Authors
```

tests/fiaas_deploy_daemon/specs/v3/data/examples/single_extra_config_map.yml
```diff
@@ -0,0 +1,18 @@
+
+# Copyright 2017-2019 The FIAAS Authors
```
gregjones, line 2:
```suggestion
# Copyright 2017-2020 The FIAAS Authors
```

gregjones (commented):
This looks very nice now 👍 
I noticed the copyright years in the new files aren't updated though, sorry to be picky. I've added the suggested changes with this review.

mortenlj:
@gregjones : This looks good now, doesn't it?

We still haven't figured out how to get Semaphore to build from forks automatically I guess, but I'll see if I can trigger something...

gregjones (approved):
👍

skogie:
@mortenlj looks like the build is failing now because some of the tasks are not running, can we try to trigger a new build and see if we can merge it?

mortenlj:
I have a feeling this PR can be closed ... ?

PR #116, opened 2020-09-25
Fix e2e tests for Kubernetes 1.12 and 1.14

### LINKED ISSUE #115: e2e tests fail because kind images for 1.12 and 1.14 have expired certificates
It looks like the CI build for #105 fails because the kind clusters used for e2e tests with Kubernetes version 1.12 and 1.14 fail to start. I looked into it and it looks like it might be the same issue with expired certificates which was fixed for Kubernetes 1.9 in #45 . Looks like we might need a similar fix for these Kubernetes versions too.

I found this in the kubelet logs from the kind docker container:
```
E0922 09:27:13.978134     405 bootstrap.go:264] Part of the existing bootstrap client certificate is expired: 2020-09-10 23:17:42 +0000 UTC
```

oyvindio:
Update Kubernetes 1.14 image to 1.14.10, which seems to work.

For Kubernetes 1.12, even the latest bsycorp/kind image seems to be built over a year ago, which means some internal certificate already has expired.  I've built a new image as fiaas/kind:v1.12.10 from bsycorp/kind@5b67b70 following the same procedure as in https://github.com/fiaas/fiaas-deploy-daemon/pull/45, and modified the e2e test for Kubernetes 1.12 to use that image.

Closes #115

### DIFF
--- a/tests/fiaas_deploy_daemon/conftest.py
+++ b/tests/fiaas_deploy_daemon/conftest.py
@@ -188,6 +188,6 @@ def _open():
         yield mock_open
 
 
-@pytest.fixture(scope="session", params=("v1.9.11", "v1.12.10", "v1.14.4", pytest.param("v1.16.3", marks=pytest.mark.e2e_latest)))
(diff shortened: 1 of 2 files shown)


gregjones (approved):
🏆

PR #117, opened 2020-09-25
app based service accounts

stribb:
Enable app-based service accounts behind a feature flag, `--enable-service-account-per-app`.


fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
```diff
@@ -82,7 +83,7 @@ def deploy(self, app_spec, selector, labels, besteffort_qos_is_required):
 
         automount_service_account_token = app_spec.admin_access
         init_containers = []
-        service_account_name = "default"
+        service_account_name = app_spec.name if self._enable_service_account_per_app is True else "default"
```
gregjones, line 86: Does this cause the service-account to be created if it doesn't exist?

sa-mao: No I don't think so. Creating the service account is not part of this PR

gregjones: Ok. The discussion in the previous PR seemed to arrive at the conclusion that's how we'd implement this. It could be an idea to have some content in the PR description explaining the context - i.e. link it to the issue, and if the implementation isn't matching the proposal, explain it.

Either way, the help text suggests that if the account doesn't exist then it will fallback to use default. Is that the kubernetes default behaviour, or will it fail the deploy?

stribb: Thanks Greg. We haven't checked that situation but I expect it would fail. I can update the help text: it was written aspirationally at the beginning of the PR!

I agree, we need to give more context: we'll make sure that we match it with a GitHub issue so you see what we're aiming for. Though right now, it feels more like sculpture: we'll find out what the feature ends up like when it's done ;)

stribb: I also wanted to shout out to @sa-mao and @PerGon as co-authors: we worked on this in a "mob session" over the last two weeks.

gregjones: Ah, ok. Sounds good.
Cc @cdemonchy as you folks were interested in this feature too

PerGon:
Thanks for the comments @gregjones . Here's some additional information - likely a good idea to add in the PR description.

This PR is the starting point for https://github.com/fiaas/fiaas-deploy-daemon/issues/72 and a different approach of https://github.com/fiaas/fiaas-deploy-daemon/pull/107

In https://github.com/fiaas/fiaas-deploy-daemon/pull/107 more details on the implementation of this feature was discussed.
The idea is that, behind a feature flag, operators can enable a service account with the same name of the app to be attached to the application.

This PR simply attaches the service account to the `Deployment` object - in a future PR, the behaviour to create the service account (or delete in the case of app deletion) will be added.

oyvindio:
> This PR simply attaches the service account to the `Deployment` object - in a future PR, the behaviour to create the service account (or delete in the case of app deletion) will be added.

It might be worth considering the behavior when the capability of FIAAS managing the per-application service account is added. If fiaas-deploy-daemon is running with this configuration and there is another system managing the service accounts which are used when the change to have fiaas-deploy-daemon manage the service account is rolled out, there might be a conflict if the two systems have differing opinions of how the service account should look. This behaviour could of course be rolled out behind another configuration flag if necessary. 
That said I I think an iterative approach is a good idea.

tests/fiaas_deploy_daemon/specs/v3/test_v3factory.py
```diff
@@ -83,6 +83,7 @@
         "ingresses[0].pathmappings[0].path": "/",
         "ingresses[0].pathmappings[0].port": 80,
         "secrets_in_environment": False,
+        ""
```
oyvindio, line 86: I might be missing something, but I don't think this has any effect in practice. Is it a typo?

tests/fiaas_deploy_daemon/deployer/kubernetes/test_deployment_deploy.py
```diff
@@ -81,12 +86,14 @@ def environment(self, request):
...
-        ("gke", {}, True, True),
+        # key: (infrastructure, global_env, set[Feature]
+        ("gke", {}, {Feature.USE_IN_MEMORY_EMPTYDIRS, Feature.DISABLE_DEPRECATED_MANAGED_ENV_VARS}),
+        ("gke-w/service-per-acct", {}, {Feature.USE_IN_MEMORY_EMPTYDIRS, Feature.DISABLE_DEPRECATED_MANAGED_ENV_VARS,
+         Feature.ENABLE_SERVICE_ACCOUNT_PER_APP}),
         ("diy", {
             'A_GLOBAL_DIGIT': '0.01',
             'A_GLOBAL_STRING': 'test',
-        }, True, True),
+        }, {Feature.USE_IN_MEMORY_EMPTYDIRS, Feature.DISABLE_DEPRECATED_MANAGED_ENV_VARS}),
```
oyvindio, line 96: This seems like a nice readability improvement 👍

fiaas_deploy_daemon/config.py
```diff
@@ -125,6 +125,10 @@
 if the application doesn't specify any.
 """
 
+ENABLE_SERVICE_ACCOUNT_PER_APP = """
+Look for a service account by app name; use this if we find one, else fall back to
+the namespace's default.
+"""
```
oyvindio, line 131: I don't think this text matches the behavior as it works in this branch. Maybe it should be changed to something like "Use a service account matching the application name".

oyvindio:
This feature was merged in #159

Issue #114, opened 2020-09-21
Support namespace-level 'issuer' for TLS


gregjones:
Currently, FIAAS supports configuring ingresses to support TLS using LetsEncrypt via the "cluster-Issuer" feature of cert-manager. To allow for custom domains using DNS challenges (necessary for provisioning valid certificates in a multi-cluster setup, where HTTP-challenge doesn't work well) we'd like to allow for users to also configure their apps to use an "issuer" instead (these exist inside the same namespace). For cert-manager, this requires the Ingress to have different annotations (certmanager.k8s.io/cluster-issuer vs. certmanager.k8s.io/issuer). Currently in FIAAS, this value is fixed, and only the value of the annotation can be changed (this allows switching between letsencrypt and letsencrypt-staging, for example).

The least intrusive way to enable this feature seems to be extending the configuration at the fiaas-deploy-daemon level, where we can introduce a mapping between domain suffixes and the key issuer, and the app config won't need to change.

The related config for TLS issuing would look something like:

```yaml
tls-certificate-issuer: letsencrypt
tls-certificate-issuer-type-overrides:
  - foo.example.com=certmanager.k8s.io/issuer
  - bar.example.com=certmanager.k8s.io/issuer
tls-certificate-issuer-type-default: certmanager.k8s.io/cluster-issuer
```

Where `tls-certificate-issuer` currently exists, and remains the same.
`tls-certificate-issuer-type-overrides` is a new key and contains a mapping between domain suffixes and the annotation to use when configuring TLS for an ingress (a domain matches if the value in the configuration is a sub-domain, that is a suffix with an additional '.'-separator; in the case of multiple matches, the ~first~ most-specific (i.e. longest) in the list will be used).
`tls-certificate-issuer-type-default` is another new key, and decides which issuer-type should be used when the domain doesn't match any configured in the previous section; the default will be to continue using `cluster-issuer`.

This feature will also require a related change in how ingresses are created: currently, the ingress annotation to configure TLS is added after the creation of multiple ingresses, and so it would currently be the case that multiple hosts, requiring different issuers, would be created using the same Ingress, which won't work correctly. So as part of this change we will also change that, so that the annotations added to configure TLS are taken into account when creating one or multiple ingresses.

This change will be completely transparent and have no impact for fiaas-admins not configuring tls-certificate-issuer-type-overrides.

PR #118, opened 2020-10-06
Refactor order of operations for hosts vs. annotations

gregjones:
This change moves the expansion of default-hosts to be before the grouping
by annotations. To support setting different annotations per-issuer for TLS,
depending on the host, requires further splitting ingress-items into
separate ingress objects.

Relates to #114


xavileon (approved):

oyvindio (approved):
Looks good to me!

PR #119, opened 2020-10-06
Support alternate issuer types for TLS

### LINKED ISSUE #114: Support namespace-level 'issuer' for TLS

gregjones:
This allows, for example, using `certmanager.k8s.io/issuer: letsencrypt` for a namespace-level issuer, instead of always cluster-issuer. To allow for a mixture of issuer types between the hosts an app might use, this requires splitting the ingress items amongst separate ingress objects, similar to how it's done for supporting separate annotations in the configuration

Requires #118

Closes #114

### DIFF
--- a/fiaas_deploy_daemon/deployer/kubernetes/ingress.py
+++ b/fiaas_deploy_daemon/deployer/kubernetes/ingress.py
@@ -41,6 +41,7 @@ class IngressDeployer(object):
         self._host_rewrite_rules = config.host_rewrite_rules
         self._ingress_tls = ingress_tls
         self._owner_references = owner_references
+        self._tls_issuer_type_default = config.tls_certificate_issuer_type_default
 
     def deploy(self, app_spec, labels):
         if self._should_have_ingress(app_spec):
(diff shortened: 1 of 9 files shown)


fiaas_deploy_daemon/deployer/kubernetes/ingress.py
```diff
@@ -77,24 +80,45 @@ def _expand_default_hosts(self, app_spec):
...
-        ''' Group the ingresses so that those with annotations are individual, keeping all without
-        annotations together
+    def _get_issuer_type(self, host):
+        for (suffix, issuer_type) in self._tls_issuer_type_overrides:
+            if host and (host == suffix or host.endswith("." + suffix)):
+                return issuer_type
+
+        return self._tls_issuer_type_default
+
+    def _group_ingresses(self, app_spec):
```
xavileon, line 90: Thinking out loud: I was wondering if the code would look simpler if we treat the certmanager annotations as regular annotations 🤔  This might mean an extra ingress for each overriden issuer though (i.e. counting towards rate limiting) and a refactor of the `apply` method to remove how the annotations are handled there. It wouldn't change how this feature behaves so probable for a later PR/refactor.

xavileon (dismissed):
Change looks good to me! 🏆

amanya (dismissed):

adriengentil (dismissed):

oyvindio (approved):
Looks good to me

xavileon (approved):

Issue #12, opened 2019-07-08
Propagation of config flags as environment variables can lead to config changes not taking effect on redeploy


oyvindio:
When fiaas-deploy-daemon deploys itself, it sets some environment variables based on the internal state of its config object (see below). Since the config object is based on data from environment variables and the `cluster_config.yaml` file being merged (in that order), this can lead to a configuration change in the cluster_config.yaml file not being picked up since a value is already set in the corresponding environment variable.

https://github.com/fiaas/fiaas-deploy-daemon/blob/4d4bf405dd0ac729292f442513f9589a9b0516fe/fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py#L223-L236

The typical way this occurs is when bootstrapping fiaas-deploy-daemon via Skipper. When this is done, the `fiaas-deploy-daemon` configmap, which usually provides the `cluster_config.yaml` file is not mounted in the Pod. This means that the default values for flags like `FIAAS_ENVIRONMENT` and `LOG_FORMAT` are used. To change these later, just changing the configmap is not enough; to make the config change actually be applied you have to change the configmap, then remove the corresponding environment variables from the fiaas-deploy-daemon deployment resource, which will trigger a redeploy that then applies the changes. 

This behavior might be improved by not propagating the state from the internal config object at all when fiaas-deploy-daemon is deploying itself

mortenlj:
It might be worth looking into if we really need this behaviour at all. If we say fdd should always read its config from file and ignore the environment, the environment variables will only be used by normal apps in the cluster.
We already have a different mechanism for setting global env-variables, where the cluster config explicitly defines the values. .

srvaroa:
+1, using always config files seems like the simplest choice to remove ambiguity.

mortenlj:
We currently load config from environment variables, config file or options on the command line. Both command line and environment variables are quite limited (string values only), while the config file is YAML and could potentially support more complex values.

We already have a few values that would be easier to express using complex values, and we are discussing others. We need to consider if we need to be able to configure all values using environment values and/or command line options, or if we should do all configuration using the config file?

oyvindio:
[...]

PR #120, opened 2020-10-06
Log a warning if it looks like environment variables are used to set config options

oyvindio:
First part of fixing #12. Checks for presence of environment variables that have the same key/value as in the configuration object, then logs a warning indicating that this mechanism for configuring fiaas-deploy-daemon is deprecated.
I put this outside the config module since the logger needs the config object to configure itself.

### DIFF
--- a/fiaas_deploy_daemon/__init__.py
+++ b/fiaas_deploy_daemon/__init__.py
@@ -17,6 +17,7 @@
 from __future__ import absolute_import
 
 import logging
+import os
 import signal
 import sys
 import threading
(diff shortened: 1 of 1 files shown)


tg90nor (dismissed):

gregjones (dismissed):

oyvindio:
Rebased and removed disable_pipeline_consumer after merging #105

gregadevinta (approved):
👍

gregjones (approved):
👍 👍

PR #121, opened 2020-10-06
Remove ability to set config options via environment variables

### LINKED ISSUE #12: Propagation of config flags as environment variables can lead to config changes not taking effect on redeploy

oyvindio:
This is an attempt at solving #12, based on the suggestion here https://github.com/fiaas/fiaas-deploy-daemon/issues/12#issuecomment-688109631. 

fiaas-deploy-daemon's configuration is based on a merge (by order of precedence) of command-line flags, environment variables and the cluster_config.yaml configuration file, typically provided via a ConfigMap. 
Because of the order of precedence (command-line flags > environment variables > config file), and because fiaas-deploy-daemon sets environment variables from its configuration when deploying itself, this could lead to situations where a change in the config file would not take effect if the configuration option that was change was shadowed by a environment variable. There is an example of this in the associated issue.

To remove the potential for confusion stemming from this behavior, remove the ability to configure fiaas-deploy-daemon via environment variables. This is a backwards incompatible change for deployments of fiaas-deploy-daemon relying on environment variables for configuration. Deployments using environment variables will need to switch to command-line flags or config file/ConfigMap before updating to a version of fiaas-deploy-daemon including this fix.

Follow-up to #120 (this changeset should remove the warning implemented there)
Closes #12

### DIFF
--- a/fiaas_deploy_daemon/__init__.py
+++ b/fiaas_deploy_daemon/__init__.py
@@ -165,9 +165,10 @@ def warn_if_env_variable_config(config, log):
     possible_config_env_variables = sorted(configuration_env_variable_keys & environ_keys)
     if len(possible_config_env_variables) > 0:
         log.warn("found configuration environment variables %s. The ability to configure fiaas-deploy-daemon via environment variables " +
-                 "will be removed. If these environment variables are the primary source for this configuration, please switch to " +
-                 "configuring via a config file/ConfigMap or command-line flags. See " +
-                 "https://github.com/fiaas/fiaas-deploy-daemon/issues/12 for more information", ', '.join(possible_config_env_variables))
+                 "has been removed. If you are trying to use these environment variables to configure fiaas-deploy-daemon, " +
+                 "that configuration will not take effect. Please switch to configuring via a config file/ConfigMap or command-line " +
+                 "flags. See https://github.com/fiaas/fiaas-deploy-daemon/issues/12 for more information.",
+                 ', '.join(possible_config_env_variables))
 
 
 def main():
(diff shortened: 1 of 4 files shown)


oyvindio:
It is ~5 months since the warning log message for environment variable configuration was added in #126, I think this should probably be OK to merge now. I've left the logging in place, but the message now states that environment variable based configuration is no longer available.

gregjones (approved):
👍

PR #123, opened 2020-10-08
Create svc account per app

sa-mao:


fiaas_deploy_daemon/deployer/kubernetes/adapter.py
```diff
@@ -29,19 +29,22 @@ class K8s(object):
     """Adapt from an AppSpec to the necessary definitions for a kubernetes cluster
     """
 
-    def __init__(self, config, service_deployer, deployment_deployer, ingress_deployer, autoscaler):
+    def __init__(self, config, service_deployer, deployment_deployer, ingress_deployer, autoscaler, service_account_deployer):
```
gregjones, line 32: With this change you'll have to update some of the tests that pass in an argument here - I think if you look, you'll see how the test is creating mocks for the other parameters already, and you can follow the same pattern for this one. A test in that class that verifies the deployer is only called when the config says it's supposed to be would be good to see.

fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
```diff
@@ -82,7 +83,7 @@ def deploy(self, app_spec, selector, labels, besteffort_qos_is_required):
 
         automount_service_account_token = app_spec.admin_access
         init_containers = []
-        service_account_name = "default"
+        service_account_name = app_spec.name if self._enable_service_account_per_app is True else "default"
```
gregjones, line 86:
```suggestion
        service_account_name = app_spec.name if self._enable_service_account_per_app else "default"
```
readability tip: Comparing booleans to values isn't necessary, you can just test them directly

fiaas_deploy_daemon/deployer/kubernetes/service_account.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+class ServiceAccountDeployer(object):
+    def __init__(self, config, owner_references):
+        self._owner_references = owner_references
+
+    def deploy(self, app_spec, labels):
+        self._create(app_spec, labels)
+
+    def delete(self, app_spec):
```
gregjones, line 39: As mentioned when we chatted, this method isn't necessary; since we're setting the ownerReferences, the deletion of dependent objects will happen automatically, and we don't need to worry whether or not we're deleting something that fiaas didn't create

fiaas_deploy_daemon/deployer/kubernetes/service_account.py
```diff
@@ -0,0 +1,81 @@
...
+            ServiceAccount.delete(app_spec.name, app_spec.namespace)
+        except NotFound:
+            pass
+
+    @retry_on_upsert_conflict
+    def _create(self, app_spec, labels):
+        LOG.info("Creating/updating serviceAccount for %s with labels: %s", app_spec.name, labels)
+        metadata = ObjectMeta()
+        image_pull_secrets = []
+        automount_service_account_token = True
```
gregjones, line 55: This is now unused

fiaas_deploy_daemon/deployer/kubernetes/service_account.py
```diff
@@ -0,0 +1,81 @@
...
+              return
+
+            ServiceAccount.delete(app_spec.name, app_spec.namespace)
+        except NotFound:
+            pass
+
+    @retry_on_upsert_conflict
+    def _create(self, app_spec, labels):
+        LOG.info("Creating/updating serviceAccount for %s with labels: %s", app_spec.name, labels)
+        metadata = ObjectMeta()
```
gregjones, line 53: This is now unused; it's initialised below

fiaas_deploy_daemon/deployer/kubernetes/service_account.py
```diff
@@ -0,0 +1,81 @@
...
+        automount_service_account_token = True
+        try:
+            service_account = ServiceAccount.get(app_spec.name, app_spec.namespace)
+            if not self._owned_by_fiaas(service_account):
+              return
+        except NotFound:
+            pass
+
+        default_service_account = ServiceAccount.get("default", app_spec.namespace)
+        image_pull_secrets = default_service_account.imagePullSecrets
```
gregjones, line 64: Is it possible that either the default service-account won't exist, or that it won't have imagePullSecrets?

fiaas_deploy_daemon/deployer/kubernetes/service_account.py
```diff
@@ -0,0 +1,81 @@
...
+        service_account_name = app_spec.name
+        custom_labels = labels
+        custom_annotations = {}
+        if "service_account" in app_spec.labels:
```
gregjones, line 68: I think this isn't correct; app_spec.labels is a namedtuple, with properties, rather than a dict. You should update LabelAndAnnotationSpec in models.py to add the new property, and then populate it in _labels_annotations_spec inside v3/factory.py
(Then the 'if' becomes unnecessary here; the value will definitely be a dict, and will be mergeable even if it's empty)

fiaas_deploy_daemon/deployer/kubernetes/service_account.py
```diff
@@ -0,0 +1,81 @@
...
+            custom_labels = merge_dicts(app_spec.labels.service_account, custom_labels)
+        if "service_account" in app_spec.annotations:
```
gregjones, line 70: (Annotations and Labels work in exactly the same way, using the same type, so the comment above applies to this as well)

setup.py
```diff
@@ -32,7 +32,7 @@ def read(filename):
     "pinject == 0.14.1",
     "six == 1.12.0",
     "dnspython == 1.16.0",
-    "k8s == 0.15.0",
+    "k8s == 0.17.0",
```
gregjones, line 35: Also as mentioned, it might be cleaner to open a new PR that just updates the library version and fixes the existing tests, so it's totally clear the new functionality doesn't impact anything (without it being enabled)

tests/fiaas_deploy_daemon/deployer/kubernetes/test_deployment_deploy.py
```diff
@@ -81,12 +86,14 @@ def environment(self, request):
...
-        ("gke", {}, True, True),
+        # key: (infrastructure, global_env, set[Feature]
+        ("gke", {}, {Feature.USE_IN_MEMORY_EMPTYDIRS, Feature.DISABLE_DEPRECATED_MANAGED_ENV_VARS}),
+        ("gke-w/service-per-acct", {}, {Feature.USE_IN_MEMORY_EMPTYDIRS, Feature.DISABLE_DEPRECATED_MANAGED_ENV_VARS,
+         Feature.ENABLE_SERVICE_ACCOUNT_PER_APP}),
         ("diy", {
             'A_GLOBAL_DIGIT': '0.01',
             'A_GLOBAL_STRING': 'test',
-        }, True, True),
+        }, {Feature.USE_IN_MEMORY_EMPTYDIRS, Feature.DISABLE_DEPRECATED_MANAGED_ENV_VARS}),
```
gregjones, line 96: This is a really nice improvement for the testing 🏆

tests/fiaas_deploy_daemon/deployer/kubernetes/test_service_account_deploy.py
```diff
@@ -0,0 +1,59 @@
...
+from fiaas_deploy_daemon.config import Configuration
+from fiaas_deploy_daemon.deployer.kubernetes.service_account import ServiceAccountDeployer
+from fiaas_deploy_daemon.specs.models import LabelAndAnnotationSpec
+
+from utils import TypeMatcher
+
+SERVICES_ACCOUNT_URI = '/api/v1/namespaces/default/serviceaccounts/'
+LABELS = {"service": "pass through"}
+
+class TestServiceDeployer(object):
```
gregjones, line 32:
```suggestion
class TestServiceAccountDeployer(object):
```

tests/fiaas_deploy_daemon/deployer/kubernetes/test_service_account_deploy.py
```diff
@@ -0,0 +1,59 @@
...
+#    def service_type(self, request):
+#        return request.param
+
+    @pytest.fixture
+    def deployer(self, owner_references):
+        config = create_autospec(Configuration([]), spec_set=True)
+        return ServiceAccountDeployer(config, owner_references)
+
+    @pytest.mark.usefixtures("get")
+    def test_deploy_new_service_account(self, deployer, post, app_spec, owner_references):
```
gregjones, line 43: With the new logic about dealing with an existing account with the app's name, and copying imagePullSecrets from default, I think a couple more tests to handle those extra cases would be good to see:
 * if the account exists and isn't owned by fiaas, verify no calls are made
 * when the default account has imagePullSecrets, they are present in the new account that's created

gregjones (commented):
This is moving well in the right direction, I've added a few comments, and as I mentioned when we chatted, I'll suggest something for the e2e tests to have it run with a different 'fdd' fixture with the new flag set

fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
sa-mao: The tests fail if we don't have a value for that I think 🤔

fiaas_deploy_daemon/deployer/kubernetes/service_account.py
sa-mao: AFAIK the default serviceAccount is always there but I will wrap in try catch anyways.

fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
sa-mao: That was before.

fiaas_deploy_daemon/deployer/kubernetes/service_account.py
sa-mao: I will keep it anyways. WDYT?

setup.py
sa-mao: It is done. check https://github.com/fiaas/fiaas-deploy-daemon/pull/128, the current PR was rebased on the top of that PR.

tests/fiaas_deploy_daemon/deployer/kubernetes/test_deployment_deploy.py
sa-mao: Kudos to @stribb 🥇

tests/fiaas_deploy_daemon/deployer/kubernetes/test_service_account_deploy.py
sa-mao: Would you be kind enough to write the later? (looking for inspiration :p)

oyvindio (commented):
Thank you for the PR, I think this feature will be an useful addition! 

I think this is starting to look good. I've left a few comments - several of them are minor/nitpicking, but the most important points are maybe these:
* It would be really useful to have tests covering the behavior when 1) propagating imagePullSecrets from the default service account and 2) using but not overwriting existing non-managed application service accounts
* I don't think the delete logic is necessary any more because of ownerReferences
* Update the command line flag documentation to match the behavior

It would also be useful to have an e2e test covering this feature. 

As I also mentioned in #128, the CI build unfortunately doesn't run e2e tests for PRs from forks (yet) (see #131), so if that isn't fixed before this ready to merge we can push it to a branch on the repo to before merging as a workaround to get a full CI build.

oyvindio:
For this to work when fiaas-deploy-daemon is deployed with RBAC privileges managed by skipper, the RBAC role skipper creates needs to be updated to allow read/list/create/update on serviceaccount resources. See https://github.com/fiaas/skipper/blob/master/helm/fiaas-skipper/templates/rbac.yaml

fiaas_deploy_daemon/config.py
```diff
@@ -122,6 +122,11 @@
 if the application doesn't specify any.
 """
 
+ENABLE_SERVICE_ACCOUNT_PER_APP = """
+Look for a service account by app name; use this if we find one, else fall back to
+the namespace's default.
+"""
```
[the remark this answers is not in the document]
sa-mao: That's a nice suggestion, I will use it :)

fiaas_deploy_daemon/deployer/kubernetes/adapter.py
```diff
@@ -52,6 +55,8 @@ def delete(self, app_spec):
         self._autoscaler_deployer.delete(app_spec)
         self._service_deployer.delete(app_spec)
         self._deployment_deployer.delete(app_spec)
+        if self._enable_service_account_per_app is True:
+            self._service_account_deployer.delete(app_spec)
```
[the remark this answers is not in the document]
sa-mao: Sure! I agree!

sa-mao:
@oyvindio I think it already does: https://github.com/fiaas/skipper/blob/master/helm/fiaas-skipper/templates/rbac.yaml#L70

j18e:
Here's the rbac vision I've got in my head so far. Could anyone take a look and let me know if I'm missing anything?

[old fiaas rbac diagram](https://docs.google.com/drawings/d/1bXP11wV6B_PEvfUGyQbmxelY6PvaHCZ4SuVOL_x9ySs/edit?usp=sharing)
[future fiaas rbac diagram](https://docs.google.com/drawings/d/12_ePogtjcdmQt6PTeN0E1Pyxn1-HbE3vmMeuIGvkecE/edit?usp=sharing)

oyvindio (commented):
What is the status of this PR, is there any progress? I took another look at this now and to me it looks like it is pretty close to ready. I tried test driving it a bit locally, and it seems like there is a typo that results in a `NameError` when trying to deploy the service account (see comment in line), but that is an easy fix. Also as I mentioned in my previous review, before merging this I think it would be very useful to have some additional tests covering the behavior for different cases when 1) propagating imagePullSecrets from the default service account if any exist, and also the cases when 2) there are existing service accounts managed by FIAAS/not managed by FIAAS.

birgirst:
> What is the status of this PR, is there any progress? I took another look at this now and to me it looks like it is pretty close to ready. 

I'm planning to pick this one up. I'll need to read this through and see if I can address some of things that have not been addressed already.

j18e:
this was addressed in https://github.com/fiaas/fiaas-deploy-daemon/pull/159

PR #127, opened 2020-10-19
Fix TLS hosts for ingress created for custom issuer

gregjones:
The TLS hosts for the ingress used when the domain is for an
overridden issuer shouldn't include the "collapsed" host, which
is based on the default-suffixes.


adriengentil (approved):
LGTM!

ferwguerra (approved):
LGTM

xavileon (approved):

PR #130, opened 2020-10-28
Fix default suffix hosts being rewritten

gregjones:
Changes in #118 resulted in a change in behaviour, by using
the rewrite patterns also on the hosts generated from the default
suffixes, which wasn't previously the case.
This change applies those rewrites only to the hosts specified by
the application, before they are expanded.

Kudos to @oyvindio for the simple breaking testcase 🏆


oyvindio (approved):
LGTM

I think one of the CI builds might have failed because of something unrelated/the test cluster just didn't come up in time. Maybe try retriggering it?

amanya (approved):

adriengentil (approved):

gregjones:
It took a few attempts to get them to pass, I wonder if something has changed with the environment to make it more flaky :/

PR #132, opened 2020-11-04
Revert "Fix default for tls-issuer-type"

gregjones:
certmanager (with no hyphen) is the previous default. Newer versions of cert-manager have changed to add the hyphen.

This reverts commit b637e8f104c535dc906b782817623fe1e02f6dfa to restore the previous default. If you're using a newer version of cert-manager, you can now change this default via the config.

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -39,7 +39,7 @@ blocks:
           - docker login --username "${DOCKER_USERNAME}" --password-stdin <<< "${DOCKER_PASSWORD}"
           - ./bin/docker_build
           # Run end-to-end/integration tests
-          - tox -e integration_test -- -n 7 --use-docker-for-e2e
+          - tox -e integration_test -- -n 4 --use-docker-for-e2e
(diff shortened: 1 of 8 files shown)


gerardsegarra (approved):

oyvindio (dismissed):

amanya (approved):

gregjones:
I tried many times getting the CI to pass, but it kept failing. I did some tests, and it seems lowering the parallelism of the e2e tests helps, so I've done that. It makes some sense that running 4 versions of kind and then using 7 cpus (out of 8) could cause an issue. Compared to other successful runs, this adds about 2 minutes to the build. I think a better solution could be to run each version's tests serially (with the tests running in parallel within that scope)... maybe the matrix option from semaphore can help with that. Or we drop some versions from the list that get tested against.

oyvindio:
Let's try running with reduced parallellism for a while to see how that works. 
It would be very nice to to reduce the number of versions we test against, as some of them are getting quite old.

oyvindio (approved):

PR #135, opened 2020-11-20
Keep autoscaler min_replicas=max_replicas if max < min

xavileon:
If the user specifies a minimum number of replicas higher than the
max number of replicas, the autoscaling will set the deployment
number of replicas to the minimum (which is higher than the maximum).

Setting max < min doesn't make much sense in any case, but this
is even more unexpected when the user specify a max=1 and doesn't
even specify minimum and it takes the default(2).

This fix makes sure autoscaler.min_replicas is never higher than
autoscaler.max_replicas.

Signed-off-by: Xavi León <xavi.leon@adevinta.com>

### DIFF
--- a/tests/fiaas_deploy_daemon/specs/v3/test_v3factory.py
+++ b/tests/fiaas_deploy_daemon/specs/v3/test_v3factory.py
@@ -95,6 +95,12 @@ TEST_DATA = {
         "autoscaler.max_replicas": 3,
         "autoscaler.cpu_threshold_percentage": 50,
     },
+    "autoscaling_max_less_than_min": {
+        "autoscaler.enabled": False,
+        "autoscaler.min_replicas": 1,
+        "autoscaler.max_replicas": 1,
+        "autoscaler.cpu_threshold_percentage": 50,
+    },
     "multiple_hosts_multiple_paths": {
         "ingresses[0].host": None,
         "ingresses[0].pathmappings[0].path": "/0noport",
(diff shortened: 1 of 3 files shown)


gregjones (commented):
👍  But I think worth a note in the docs explaining this https://github.com/fiaas/fiaas-deploy-daemon/blob/master/docs/v3_spec.md#minimum

gregjones (approved):
👍

tests/fiaas_deploy_daemon/specs/v3/test_v3factory.py
```diff
@@ -372,6 +378,15 @@ def test_name_and_image(self, load_app_config_testdata, factory, filename):
...
 
+    @pytest.mark.parametrize("filename", (
+            "autoscaling_max_less_than_min",
+    ))
+    def test_autoscaling_max_less_than_min(self, load_app_config_testdata, factory, filename):
+        app_spec = factory(UID, NAME, IMAGE, load_app_config_testdata(filename), ["IO"], ["foo"], "deployment_id", NAMESPACE,
+                           None, None)
+        assert app_spec.autoscaler.min_replicas == TEST_DATA[filename]["autoscaler.min_replicas"]
+        assert app_spec.autoscaler.max_replicas == TEST_DATA[filename]["autoscaler.max_replicas"]
+
```
oyvindio, line 389: Tests are auto-generated from `TEST_DATA` using `pytest_generate_tests` and this test function https://github.com/fiaas/fiaas-deploy-daemon/blob/0b32bac1158e64a932cccd84dd646924940fa56d/tests/fiaas_deploy_daemon/specs/v3/test_v3factory.py#L444-L452 The generated tests will assert that the resulting app_spec from the file with the name in the `TEST_DATA` key have the defined key+values under that key, so I think this test case is actually already covered by a generated test just by having the entry in `TEST_DATA`. You can try to  remove this test function and remove/re-add the change in factory.py to verify that the generated test breaks

oyvindio (dismissed):
This change makes a lot of sense, I see how the described behavior is confusing. See comment about possibly redundant test case, but otherwise this looks good to me 👍

xavileon:
@oyvindio @mortenlj I completely forgot about this one. I've rebased and addressed the duplicated test and now it seems in better shape.

failing build (removing fix and redundant test) -> https://fiaas-svc.semaphoreci.com/jobs/3775f9bc-5aeb-42cb-82c3-c1feab1683d4
passing build (adding the fix back) -> https://fiaas-svc.semaphoreci.com/jobs/dfb85693-31b4-4436-8cc6-40295fdcd590#L641

xavileon:
@fiaas/maintainers PTAL

oyvindio (approved):
Thanks for updating the PR, it looks good to me 👍
I forgot about following this up with a review too; thank you for your patience

xavileon:
No worries :) I was not the fastest updating it either after the comments. Thanks!

PR #128, opened 2020-10-19
Upgrade k8s library to 0.17.0

sa-mao:
Upgrade k8s Library to 0.17.0 in order to introduce the ServiceAccount class.

PR #136, opened 2021-01-05
DNM - Upgrade k8s library to 0.17.0 (internal branch of fork)

PerGon:
This is the same exact code as https://github.com/fiaas/fiaas-deploy-daemon/pull/128 - but rebased and done in a branch instead of a fork. It's done in a branch to validate the CI builds.

If the CI is green in this PR, I suggest we close this PR and merge the #128 in order to maintain the proper contribution author.


oyvindio:
I took a look at the CI build here, and I haven't seen that docker engine error before. I tried re-triggering the build, and it seems to have moved past what failed previously, so maybe it was some kind of intermittent issue. This time the build failed because of e2e tests failing against Kubernetes 1.16, but this also fails on master now, so it looks like it might be an issue similar to #115, just for the v1.16.3 image. I've created https://github.com/fiaas/fiaas-deploy-daemon/pull/137 to fix the e2e test issue for Kubernetes 1.16

PerGon:
I've ran these tests multiple time locally and they come up all green.
The last commit to master (which is from a PR that was green in the tests), failed. This makes me suspect there's some flakiness in the tests.

What do you think?

oyvindio:
Yeah, I've been trying to get the master build to work properly the past few days - the e2e tests are definitely flaky in the CI build. It might be necessary to reduce the parallelism some more - I've just opened #138 to try to improve that.

PerGon:
This PR is now green. This PR is the same as #128 but in a branch instead of a fork. I would prefer #128 to be merged instead of this - hence I requested the author to rebase. I'll close this PR now.

PR #140, opened 2021-03-02
fix: Ensure the default host has a path defined

adriengentil:
Fix the case when the default ingress is wrongly created when only annotated
hosts are specified. For example, a fiaas spec like this one:
```
application: myapp
config:
    ingress:
    - host: myhost
      paths:
      - path: /
      annotations:
```
leads to the create the default ingress as follow:
```
spec:
  rules:
  - host: myapp.192.168.49.2.xip.io
```
This commit ensure that the default ingress is always created with the default
path.

### DIFF
--- a/fiaas_deploy_daemon/deployer/kubernetes/ingress.py
+++ b/fiaas_deploy_daemon/deployer/kubernetes/ingress.py
@@ -27,7 +27,7 @@ from k8s.models.common import ObjectMeta
 from k8s.models.ingress import Ingress, IngressSpec, IngressRule, HTTPIngressRuleValue, HTTPIngressPath, IngressBackend, \
     IngressTLS
 
-from fiaas_deploy_daemon.specs.models import IngressItemSpec, IngressPathMappingSpec
+from fiaas_deploy_daemon.specs.models import IngressItemSpec
 from fiaas_deploy_daemon.retry import retry_on_upsert_conflict
 from fiaas_deploy_daemon.tools import merge_dicts
 from collections import namedtuple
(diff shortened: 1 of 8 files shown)


oyvindio:
Do you think you could push the virtualenv fix as a separate PR? It would be nice to get that into master

adriengentil:
Yes, will do it!

fiaas_deploy_daemon/deployer/kubernetes/ingress.py
```diff
@@ -77,6 +77,10 @@ def _create(self, app_spec, labels):
     def _expand_default_hosts(self, app_spec):
         all_pathmappings = list(_deduplicate_in_order(chain.from_iterable(ingress_item.pathmappings
                                                       for ingress_item in app_spec.ingresses if not ingress_item.annotations)))
+
+        if not all_pathmappings:
+            all_pathmappings = [IngressPathMappingSpec(path='/', port=80)]
```
gregjones, line 82: I think this should be looking up the number for the 'http' port, the same way as it's done in the v3 Factory._ingress_items, instead of fixing 80. So maybe that method needs to be extracted. Or a `defaultPathMapping` gets added somewhere during that part of the process...?

adriengentil: So, I managed to expose an app spec that contains the default values. Can you have a look?

gregjones: I don't think the defaults are exactly what's wanted, it will depend on what app is defining as its 'http' port. So the 'raw' default for the path-spec is (path: /, port: 'http'), and during the building of the AppSpec the factory turns that 'http' into the correct port: the default for the port with name 'http' is 80, but if that's changed, then I think the default path-spec should follow it.

Maybe repetitive, but for clarity, with an app defining:

```
ports: 
  - protocol: http
    name: http
    port: 120
    target_port: 8080
```
Then the default path-spec should end up as (path: /, port: 120).

Does it make sense?

adriengentil: Yes, thank you! :facepalm: 

We already have the `PortSpec` in the `app_spec`, we can retrieve the information from there.

gregjones (approved):

xavileon (approved):

PR #141, opened 2021-03-02
Fix continuous integration

adriengentil:
Upgrade virtualenv to run the tests using tox.
Upgrade setuptools in the docker image used to build the wheels.

I don't really understand why it fails now though.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,6 +28,10 @@ RUN apk --no-cache add \
 COPY . /fiaas-deploy-daemon
 COPY .wheel_cache/*.whl /links/
 WORKDIR /fiaas-deploy-daemon
+
+# setuptools needs to be upgraded in order to perform the build of the wheels
+# Note: setuptools 45.0.0 will drop the support for python 2
+RUN pip install -U setuptools==44.1.1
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -28,6 +28,7 @@ RUN apk --no-cache add \
 COPY . /fiaas-deploy-daemon
 COPY .wheel_cache/*.whl /links/
 WORKDIR /fiaas-deploy-daemon
+RUN pip install -U setuptools
```
adriengentil, line 31: maybe there is a better way?

oyvindio: Maybe it is a good idea to pin the setuptools version ? (and possibly also include a comment for the future about why this is necessary?) It looks like setuptools [45.0.0 drops python2 support](https://setuptools.readthedocs.io/en/latest/history.html#v45-0-0), so [44.1.1](https://pypi.org/project/setuptools/44.1.1/) should be the most recent setuptools version that supports python2.

adriengentil: it makes totally sense, I pinned the version and added some comments

gregjones (approved):

oyvindio (approved):

PR #147, opened 2021-03-29
Run e2e tests on Kubernetes v1.18

tg90nor:
We are upgrading to k8s v1.18 in Finn, and would like to have e2e testing for it. From my local runs it looks like everything should work.


oyvindio (approved):
👍

gregjones (approved):
👍

PR #151, opened 2021-04-15
Document setuptools requirement in getting started instructions

oyvindio:
For the dependencies declared in [`setup_requires`](https://github.com/fiaas/fiaas-deploy-daemon/blob/master/setup.py#L92) to be installed correctly it is necessary to have a updated setuptools version, so recommend installing the latest version which still has python 2 support.

As a sidenote, after upgrading to python 3 it might make sense to look into using a pyproject.toml to replace(?)/in addition to setup.py, since that should make it possible to declare requirements such as python version and build tool version, which might help avoid running into similar issues in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,9 +42,8 @@ See the config-module for more information.
 Getting started with developing
 -------------------------------
 
-First ensure you have the most recent version of setuptools for python2. This is required for the following steps to work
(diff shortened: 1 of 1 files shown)


gregjones (approved):
👍

henrik242 (approved):

Issue #143, opened 2021-03-05
Expose fiaas-deploy-daemon version via a metric


oyvindio:
To make it easier to track which version of fiaas-deploy-daemon is running at a given time (and changes over time), expose the version (`config.version`) in a metric.

oyvindio:
Fixed in #152

PR #152, opened 2021-04-28
Expose fdd version as metric

ninaangelvik:


fiaas_deploy_daemon/deployer/version_exporter.py
```diff
@@ -0,0 +1,25 @@
...
+# limitations under the License.
+from prometheus_client import Gauge
+
+
+class VersionExporter(object):
+    """ Exports the currently running FDD version """
+    version_gauge = Gauge("deployer_fdd_version", "Current FDD version",["app_name", "version"])
+
+    def fdd_version(self, app_spec):
+        self.version_gauge.labels(app_spec.name, app_spec.version).set(1)
```
mortenlj, line 25: This will create a metric with the name and version of the deployed apps as labels, not the version of FDD that is running.

fiaas_deploy_daemon/deployer/version_exporter.py
```diff
@@ -0,0 +1,25 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from prometheus_client import Gauge
+
+
+class VersionExporter(object):
+    """ Exports the currently running FDD version """
+    version_gauge = Gauge("deployer_fdd_version", "Current FDD version",["app_name", "version"])
```
oyvindio, line 22: Maybe [Info](https://github.com/prometheus/client_python/#info) is a more suitable metric type for static information such as a version number ?

oyvindio (changes requested):
It looks like the `deployer_fdd_version` metric here effectively tracks `app_spec.version` (i.e. the version field in fiaas.yml) of the most recently deployed application. To solve #143 it might make sense to use `config.version`, which should contain the image tag for the running fiaas-deploy-daemon docker image (via the VERSION or FIAAS_VERSION env variables). This could simplify the implementation too; it might make sense to put all of it in [`Main`](https://github.com/fiaas/fiaas-deploy-daemon/blob/master/fiaas_deploy_daemon/__init__.py#L84).

oyvindio (approved):

PR #153, opened 2021-05-07
Look up ApplicationStatus using labels and generate new ApplicationStatus names with random suffix

oyvindio:
- Look up existing ApplicationStatus resources via `app` and `fiaas/deployment_id` label values.
- Create random name suffixes for new ApplicationStatus resources via `.metadata.generateName`.

Note: Using `generateName` means the prefix (value of generateName) is truncated to fit prefix + generated suffix into 63 characters if the value of generateName is too long. I don't think this matters in practice since there shouldn't be anything which depends on the name of the ApplicationStatus after this change. Application names are already effectively limited to < 63 characters because they are used in various places which must conform to DNS label format.

Background;
ApplicationStatus names are generated by suffixing the app_name with a hash of the deployment_id. This
generates a "stable" name based on app_name and deployment_id, so that fiaas-deploy-daemon can look up a
ApplicationStatus by name as long as it has these values later.  The builtin `hash()` is used to create these
hashes. In python3 the algorithm used in the `hash()` function has changed, and uses a random seed generated at
interpreter start time, so it can no longer be used in this way.

This changes fiaas-deploy-daemon to look up ApplicationStatus resources by the value of their `app` and
`fiaas/deployment_id` labels. Because of this change, it is no longer necessary to generate stable names for new
ApplicationStatus resources. Since this is no longer a requirement, use the generateName field to get Kubernetes to
generate suffixes for new ApplicationStatus resources.

This is in preparation for running fiaas-deploy-daemon on python3 (#133), so that python3 fiaas-deploy-daemon can find
ApplicationStatus resources created with python2 fiaas-deploy-daemon and the other way around. This should help avoid
unnecessary redeploys of applications, and might make the first version of fiaas-deploy-daemon on python3 easier/safer
to test in existing environments.


henrik242 (approved):

gregjones (approved):
Very nice 👍

mortenlj (approved):

PR #155, opened 2021-05-21
Revert to generating applicationstatus names using `hash` and look up applicationstatuses by name/get

oyvindio:
Because of issues with applicationstatuses look up using list with label selector being slow in some environments, I'd like to revert #153 and take a different approach and use [py27hash](https://pypi.org/project/py27hash/) to replace the python2 `hash` function for creating applicationstatus names when upgrading fiaas-deploy-daemon to python3. #153 was intended to prepare for upgrading to python3 by removing use of the `hash` function for generating names, since the implementation of that function has changed in python3.

More details about issues with slow applicationstatus lookups:
In some environments the change introduced in #153 can lead to a significant delay before fiaas-deploy-daemon starts processing new deploys after starting up. This seems to be because the change in application-status lookup api operation from get by name to list with selector can lead to the lookup being significantly slower in some namespaces and/or clusters. During normal processing of deploys this is likely not an issue, but when fiaas-deploy-daemon receives the full list of existing applications initially during the the application watch api operation just after starting up, it has to check if it should deploy any of these, so it looks up application-status resources for each. This means there is a delay of `number of applications * application-status lookup latency` before fiaas-deploy-daemon can start handling new watch events/deploys. The list latency of the application-status with selector api operation is likely dependent on how the apiserver and backing etcd cluster are scaled, and on how many application-status resources there are in the namespace and/or cluster. 

As this reverts #153 , the diff should be the inverse of that.


henrik242 (approved):

gregjones (approved):
👍

PR #164, opened 2021-11-24
switch to official kind

j18e:
using bsycorp/kind to run kubernetes clusters in docker during integration tests has caused us many hours of pain. The issue usually observed is that the Kubernetes control plan certificates which are prebaked into the kind images expire after one year, forcing us to build/push our own with their own certificates which expire after one year.

kindest/node functions well and is maintained. We should be using this instead, thereby avoiding troubleshooting issues with bsycorp's kind.

### DIFF
--- a/tests/fiaas_deploy_daemon/conftest.py
+++ b/tests/fiaas_deploy_daemon/conftest.py
@@ -191,11 +191,11 @@ def _open():
 
 @pytest.fixture(scope="session", params=(
     "v1.15.12",
-    "v1.16.13",
-    "v1.18.6",
-    "v1.19.4",
+    "v1.16.15",
+    "v1.18.19",
+    "v1.19.11",
     "v1.20.7",
-    pytest.param("v1.21.1", marks=pytest.mark.e2e_latest)
+    pytest.param("v1.21.2", marks=pytest.mark.e2e_latest)
 ))
 def k8s_version(request):
     yield request.param
--- a/tests/fiaas_deploy_daemon/utils.py
+++ b/tests/fiaas_deploy_daemon/utils.py
@@ -19,16 +19,16 @@ import contextlib
 import os
 import re
 import socket
+import subprocess
 import sys
 import tempfile
 import time
 import traceback
 from copy import deepcopy
-from collections import defaultdict
 from datetime import datetime
 from urlparse import urljoin
 
-import docker
+import json
 import pytest
 import requests
 import yaml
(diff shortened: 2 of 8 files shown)


tests/fiaas_deploy_daemon/utils.py
```diff
@@ -80,11 +79,14 @@ def crd_available(kubernetes, timeout=5):
     session.cert = (kubernetes["client-cert"], kubernetes["client-key"])
 
     def _crd_available():
-        plog("Checking if CRDs are available")
         for url in (app_url, status_url):
-            plog("Checking %s" % url)
+            plog("Check if CRDs are available at %s" % url)
```
xamebax, line 83: Nitpick: "Checking" makes more sense I think.

tests/fiaas_deploy_daemon/conftest.py
```diff
@@ -191,11 +191,11 @@ def _mock_namespace_file_open(name, *args, **kwargs):
...
     "v1.15.12",
-    "v1.16.13",
-    "v1.18.6",
-    "v1.19.4",
+    "v1.16.15",
+    "v1.18.19",
+    "v1.19.11",
     "v1.20.7",
-    pytest.param("v1.21.1", marks=pytest.mark.e2e_latest)
+    pytest.param("v1.21.2", marks=pytest.mark.e2e_latest)
```
xamebax, line 198: Is there a reason for bundling changes in these versions together with the switch?

xamebax (commented):
Nice piece of work! 👍🏼 It makes sense to me as far as I can tell, but I have two small comments in-line.

tests/fiaas_deploy_daemon/conftest.py
j18e: this just has to do with the docker images available in official kind versus bsycorp's kind

xamebax (approved):
I ran the integration tests with `tox` to double-check and approve this PR, but close to the end I got a timeout/connection refused error from `kind`.

This is most probably my machine, and I'd say that since tests pass for you locally and on semaphore, you can merge this PR.

oyvindio (changes requested):
Nice work on this! 👍 Switching to the official `kind` tool should help avoid some of the issues we've seen in the e2e tests earlier such as the certificate issue you mentioned, and hopefully make the tests a bit more deterministic also.

I think this looks mostly good, but I've left a few comments and questions in line 

I've run the tests locally a few times , and I've seen some flakiness which seems to be because when creating Application and ApplicationStatus CRDs, fiaas-deploy-daemon receives a 409 Conflict response. This causes the watcher thread to exit with an exception before it can properly start, which means fiaas-deploy-daemon doesn't work at all and eventually the test fails because its health check fails.
I think adding @retry_on_upsert_conflict on [the crd create function](https://github.com/fiaas/fiaas-deploy-daemon/blob/use-official-kind/fiaas_deploy_daemon/crd/watcher.py#L71) to enable retries on that condition should be safe (it is done for create/update of all other resources) and help make CRD creation more robust not only for the e2e tests but also in production.

oyvindio (approved):
LGTM 👍

PR #166, opened 2021-12-06
Use fiaas/k8s built-in in-cluster configuration to support time-based/refreshing service account tokens (BoundServiceAccountTokenVolume)

oyvindio:
This needs https://github.com/fiaas/k8s/pull/104 before merging.

This is mainly to support regularly re-reading the service account token
from file when in-cluster configuration is used.

This also introduces the option to explicitly override the default
in-cluster configuration by setting the api-token or api-cert config
flags. Currently these flags are ignored if
`/var/run/secrets/kubernetes.io/serviceaccount/token` exists, which
could be surprising behaviour.

If neither api-token or client-cert is set, and in-cluster configuration
can not be set up (service account token can not be read), log a warning
indicating potentially missing apiserver authentication configuration.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,7 @@ GENERIC_REQ = [
     "decorator < 5.0.0",  # 5.0.0 and later drops py2 support (transitive dep from pinject)
     "six == 1.12.0",
     "dnspython == 1.16.0",
-    "k8s == 0.17.0",
+    "k8s == 0.20.0",
     "monotonic == 1.5",
     "appdirs == 1.4.3",
     "requests-toolbelt == 0.9.1",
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -33,7 +33,7 @@ def read(filename):
     "decorator < 5.0.0",  # 5.0.0 and later drops py2 support (transitive dep from pinject)
     "six == 1.12.0",
     "dnspython == 1.16.0",
-    "k8s == 0.17.0",
+    "k8s == 0.20.0",
```
j18e, line 36: since v0.21.0 was just released, do we maybe want to use that?

tg90nor: Not needed for this pr. But we'll be going for .21 for CRD support

tg90nor (approved):
LGTM

setup.py
oyvindio: I think it is fine to upgrade to k8s 0.21.0 when taking the new models for ingress and crd into use

PR #167, opened 2022-01-10
Forward $DOCKER_HOST environment variable when running tests in tox

oyvindio:
This enables running e2e tests with docker engine running in a VM (or elsewhere), which is useful for example when using something else than Docker.app to run docker on macOS.


mortenlj (approved):
:+1:

tg90nor (approved):
👍

PR #168, opened 2022-01-13
Add CodeQL scanning workflow

emilva:
This enables CodeQL scanning for the repo.


oyvindio (approved):
I think this seems useful, so LGTM

Issue #163, opened 2021-11-17
RFC: Semantic versioning based release process


oyvindio:
# Background
The current "continuous delivery"/rolling release like workflow for releasing fiaas-deploy-daemon works by automatically having successful CI builds of the master branch update the [`latest` tag](https://github.com/fiaas/releases/blob/master/fiaas-deploy-daemon/latest.json), and have the [`stable` tag](https://github.com/fiaas/releases/blob/master/fiaas-deploy-daemon/stable.json) updated by a separate manual promotion stage in the CI tool. The tags themselves are json files contained within the [fiaas/releases](https://github.com/fiaas/releases) repository. Skipper can then be configured to deploy the version of fiaas-deploy-daemon each of these tags point to. Since https://github.com/fiaas/skipper/pull/119, skipper also supports specifying the version of fiaas-deploy-daemon as a configuration flag in its helm chart.

This workflow originates from when FIAAS was still an internal project, and was intended to make it easy to roll out new releases of fiaas-deploy-daemon in several clusters.
In an open source context however I think there are some drawbacks to this model;
- The `latest` tag should be kept functional, but every pull request merge is effectively making a release. Because of this it is difficult to test a new release before creating it, beyond what unit/e2e tests can cover, since a container image is not produced until merging to master.
- The current release model's lacks a clear cut-off point. This makes it more difficult to release backwards incompatible changes, such as removal of deprecated features, or support for old Kubernetes versions. This results in this type of change taking longer to roll out.
- There is an implicit social contract to check off in the #fiaas Slack channel before promoting a version to the stable tag. This can mean longer lead times before changes are available to roll out in all clusters as changes may require testing by other operators before promoting to the stable tag.

# Goals
- Make it easier to release changes that introduce some level of backward incompatibility, such as removing support for older Kubernetes versions.
- Reduce lead time for merging pull requests
  - Currently merging to master means updating the latest tag, which can affect other FIAAS operators
- Reduce lead time from merging pull request to having changes in production
  - If using the `stable` tag in production (as is recommended), it can take some time before that tag can be promoted, because other operators might need time to test (see below)
- Reduce need to synchronize/check off with other operators when shipping a new version of fiaas-deploy-daemon.

# Suggested Process
[...]

PR #169, opened 2022-01-17
Implement optional support for apiextensions/v1 CRD

tg90nor:
Enables support for apiextensions/v1 CustomResourceDefinitions through the optional `--use-apiextensionsv1-crd` config flag.
apiextensions/v1 CRD was introduced in Kubernetes 1.16 and apiextensions/v1beta1 CRD was removed in Kubernetes 1.22


oyvindio (changes requested):
I think this looks quite good. There seems to be some issue with the e2e tests for 1.22 that needs resolving before merging. As a minor point, I think it might be possible to reduce some duplication in the watcher logic (see inline comment).

fiaas_deploy_daemon/crd/apiextensionsv1_crd_watcher.py
```diff
@@ -0,0 +1,220 @@
...
+                self._handle_watch_event(event)
+        except NotFound:
+            self.create_custom_resource_definitions()
+        except Exception:
+            LOG.exception("Error while watching for changes on FiaasApplications")
+
+    @classmethod
+    def create_custom_resource_definitions(cls):
+        object_with_unknown_fields = {"type": "object", "x-kubernetes-preserve-unknown-fields": True}
+        application_schema_properties = {
```
oyvindio, line 70: I tried testing this a bit locally now, and I think the `spec` top level object is missing here, i.e. all of these fields should be under `spec`

oyvindio:
Passing e2e tests on Kubernetes 1.22 will also require #165, since extensions/v1beta ingress isn't available in 1.22

fiaas_deploy_daemon/crd/apiextensionsv1_crd_watcher.py
tg90nor: I indeed missed that.

fiaas_deploy_daemon/crd/apiextensionsv1_crd_watcher.py
```diff
@@ -0,0 +1,220 @@
...
+                self._watch(namespace=None)
+            else:
+                self._watch(namespace=self.namespace)
+
+    def _watch(self, namespace):
+        try:
+            for event in self._watcher.watch(namespace=namespace):
+                self._handle_watch_event(event)
+        except NotFound:
+            self.create_custom_resource_definitions()
```
oyvindio, line 63: When upgrading fiaas-deploy-daemon to this version and enabling the `--use-apiextensionsv1-crd` flag in an existing cluster, the CRDs won't be updated, since they already exist as apiextensions/v1beta1 CRDs and are served via the apiextensions/v1 api also. It seems that skipper actually always updates the CRDs when it starts (see [1] and [2]), so fiaas-deploy-daemon perhaps should do the same? 

Out of necessity CRDs are managed in both skipper an fiaas-deploy-daemon. This means that operators must make sure to upgrade and enable CRD v1 in both skipper and fiaas-deploy-daemon at the same time, otherwise these will try to enforce different versions of the CRDs. When skipper is no longer in use (#163) fiaas-deploy-daemon (only) will own the CRDs, which will remove the potential for this situation.

[1]: https://github.com/fiaas/skipper/blob/8741ab0d8b25ac5c65bf21407f991dcdb3f31cac/fiaas_skipper/__init__.py#L120
[2]: https://github.com/fiaas/skipper/blob/8741ab0d8b25ac5c65bf21407f991dcdb3f31cac/fiaas_skipper/deploy/crd/bootstrap.py#L46

fiaas_deploy_daemon/crd/apiextensionsv1_crd_watcher.py
```diff
@@ -0,0 +1,220 @@
...
+            group=group,
+            names=names,
+            versions=[version_v1],
+            preserveUnknownFields=False,
+            scope="Namespaced",
+            conversion=CustomResourceConversion(strategy="None")
+        )
+        definition = CustomResourceDefinition.get_or_create(metadata=metadata, spec=spec)
+        definition.save()
+        LOG.info("Created CustomResourceDefinition with name %s", name)
```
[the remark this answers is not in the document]
tg90nor: Done

fiaas_deploy_daemon/crd/apiextensionsv1_crd_watcher.py
tg90nor: Changed it to create/update CRD resources every time a CRD Watcher is started

tests/fiaas_deploy_daemon/crd/test_crd_watcher.py
```diff
@@ -100,7 +128,157 @@ def status_get(self):
             m.side_effect = NotFound
             yield m
 
-    def test_creates_custom_resource_definition_if_not_exists_when_watching_it(self, get, post, crd_watcher, watcher):
+    def test_creates_apiextensions_v1_crd_if_not_exists_when_watching_it(self, get, post, apiextensions_v1_crd_watcher_with_syncer,
```
oyvindio, line 131: Since CRD create/update now happens in CrdResourcesSyncerApiextensionsV1 and CrdResourcesSyncerApiextensionsV1Beta1, what do you think about changing this test and `test_creates_apiextensions_v1beta1_crd_if_not_exists_when_watching_it` to test only the `CrdResourcesSyncer$version.update_crd_resources` and move the tests into separate files for each class/api version? I think it could make the tests a little simpler and would remove the coupling between CrdWatcher and CrdResourcesSyncer in the tests - CrdWatcher tests could use a mock CrdResourcesSyncer and just assert `update_crd_resources` is called when `_watch` is called

oyvindio: It would also be nice to add a test case for when the CRDs already exist, as this PR changes the behavior to always updating the CRDs

oyvindio (commented):
I've left a few suggestions as in line comments, other than that I think this looks good

oyvindio (approved):
I think this looks good 👍  just needs the CI build to pass

oyvindio (approved):

PR #170, opened 2022-01-20
Keep Datadog sidecar alive while main container is too

biochimia:
While the main application is given a grace period to terminate, and has
the termination signal delayed using a preStop action, the datadog
sidecar would be immediately terminated. This would lead to a loss of
metrics from before the pod received the termination signal, and to the
point it actually exited.

~With this change, the full grace period for the main container is
extended to the sidecar, allowing the main container to continue
producing metrics through its very end.~

With this change, the application's pre-stop delay is extended to the
sidecar, with an additional 5 seconds to give the main container a
chance to exit gracefully while metrics are still being collected. No
guarantees, regarding metrics collection, are offered to applications
that take longer to exit.

### DIFF
--- a/fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
+++ b/fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
@@ -42,15 +42,16 @@ class DeploymentDeployer(object):
         self._secrets = deployment_secrets
         self._owner_references = owner_references
         self._extension_hook = extension_hook
+        self._pre_stop_delay = config.pre_stop_delay
         self._legacy_fiaas_env = _build_fiaas_env(config)
         self._global_env = _build_global_env(config.global_env)
         self._lifecycle = None
         self._grace_period = self.MINIMUM_GRACE_PERIOD
         self._use_in_memory_emptydirs = config.use_in_memory_emptydirs
-        if config.pre_stop_delay > 0:
+        if self._pre_stop_delay > 0:
             self._lifecycle = Lifecycle(preStop=Handler(
-                _exec=ExecAction(command=["sleep", str(config.pre_stop_delay)])))
-            self._grace_period += config.pre_stop_delay
+                _exec=ExecAction(command=["sleep", str(self._pre_stop_delay)])))
+            self._grace_period += self._pre_stop_delay
         self._max_surge = config.deployment_max_surge
         self._max_unavailable = config.deployment_max_unavailable
         self._disable_deprecated_managed_env_vars = config.disable_deprecated_managed_env_vars
@@ -121,7 +122,7 @@ class DeploymentDeployer(object):
                               strategy=deployment_strategy)
 
         deployment = Deployment.get_or_create(metadata=metadata, spec=spec)
-        self._datadog.apply(deployment, app_spec, besteffort_qos_is_required)
+        self._datadog.apply(deployment, app_spec, besteffort_qos_is_required, self._pre_stop_delay + 5)
         self._prometheus.apply(deployment, app_spec)
         self._secrets.apply(deployment, app_spec)
         self._owner_references.apply(deployment, app_spec)
--- a/fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
+++ b/fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
@@ -35,6 +35,7 @@ LOG = logging.getLogger(__name__)
 
 class DeploymentDeployer(object):
     MINIMUM_GRACE_PERIOD = 30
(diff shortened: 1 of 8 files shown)


fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py
```diff
@@ -121,7 +121,7 @@ def deploy(self, app_spec, selector, labels, besteffort_qos_is_required):
                               strategy=deployment_strategy)
 
         deployment = Deployment.get_or_create(metadata=metadata, spec=spec)
-        self._datadog.apply(deployment, app_spec, besteffort_qos_is_required)
+        self._datadog.apply(deployment, app_spec, besteffort_qos_is_required, self._grace_period)
```
xavileon, line 124: Is it possible to use `self._lifecycle` from https://github.com/fiaas/fiaas-deploy-daemon/blob/master/fiaas_deploy_daemon/deployer/kubernetes/deployment/deployer.py#L51 instead of `self._grace_period` so we ensure the same pre-stop is being used in the main and sidecar containers?

This way the datadog container will also comply with the 30 seconds minimum grace period for a graceful shutdown of the container if necessary after the SIGTERM (and before SIGKILL).

[This](https://cloud.google.com/blog/products/containers-kubernetes/kubernetes-best-practices-terminating-with-grace) is a nice short writeup on how this process works.

biochimia: The point of using `self._grace_period` here is to ensure that the datadog side car stays alive for as long as the main application container itself is.

Using `self._lifecycle`, as you suggest, means that both the datadog sidecar and the main application container would receive `SIGTERM` at the same time. From my limited experience, the datadog container exits within 1s of receiving `SIGTERM`, while the main application container could be running for another 30s before receiving a final `SIGKILL` and exiting. 

So, the question really is this: how much effort do we want to put into ensuring metrics emitted by the main container are preserved?

If we don't care about preserving metrics from the moment `SIGTERM` is issued, then `self._lifecycle` could be shared between the containers.

If, on the other hand, we want to do everything we can to preserve emitted metrics, then I think we want to take the full grace period into account, and keep it as I have it here.

mortenlj: When using the grace period like this, you force all pods using datadog to wait out their grace period before shutting down. This will make deployment way slower than it needs to be. The pod doesn't stop before all containers have exited.

biochimia: Instead of giving the datadog sidecar the full grace period, how would you feel about giving the datadog sidecar a few more seconds? Say, we'd wait `pre_stop_delay + 5` seconds before sending the sidecar a `SIGTERM`?

That would allow well behaved applications that exit quickly an opportunity to send metrics on termination. It would offer no guarantees to applications that otherwise take "too long" to exit.

xavileon: +1 to Joao's last comment. I think it's a reasonable trade-off to wait 5s after pre-stop.  K8s 1.18 introduces sidecar lifecycle hooks and ordering, so we can iterate/improve on this problem later.

biochimia: I've updated my changes so that the datadog sidecar gets a pre-stop delay based on the main container's value with the proposed additional 5 seconds.

Let me know what you think.

herodes1991:
[](https://github.com/biochimia)/sem-approve

herodes1991 (approved):
From our side looks well, can you review it again @mortenlj? Thank you!

amanya (approved):

oyvindio (commented):
Based on the problem description this seems reasonable enough to me. We don't use datadog though, so I haven't seen this issue in practice.

Using an additional pre-stop delay of 5 seconds would slow down rolling upgrades a lot less than using the full `_grace_period`, so that seems like the better approach. There is still an additional delay; if that becomes a problem it might be worth considering whether the datadog pre-stop delay should be configurable via a config flag. (although a flag to configure the delay could be added later as well).

herodes1991:
@oyvindio If I understand you correctly, are you agree with merging this pull request (that is using the pre-stop delay instead of the _grace_period) and add the config flag in next iterations?

herodes1991 (approved):

herodes1991:
@oyvindio We create a configuration flag in fiaas-deploy-daemon to activate this Lifecycle in datadog. Is this okay?

oyvindio (approved):
Thanks for adding the config flag, LGTM 👍

amanya (approved):

Issue #174, opened 2022-03-03
CrdWatcher change of behaviour


herodes1991:
When the fiaas-deploy-daemon starts running, it is always updating the crd, even if is not needed.

Before it attempted first to watch out for the events in the namespace, and then, if it received a NotFound, it created the CRD. Now it is always updating the crd event before starting watching, firing some errors in our current system.

oyvindio:
The behavior of updating the CRD every time fiaas-deploy-daemon starts was introduced in https://github.com/fiaas/fiaas-deploy-daemon/pull/169, to make it possible to upgrade to the v1 CRD API in a cluster where the application and applicationstatus already exists, by enabling the `use-apiextensionsv1-crd` config flag, and also make it possible downgrade to the v1beta1 CRD API by disabling the same flag later if necessary. With the previous behavior, this would be difficult to do without updating the application and applicationstatus CRDs manually. 
What kind of errors do you see as a result of this?

herodes1991:
In our system, the team here managing fiaas isn't the same as the one that manages the clusters, meaning that fiaas-deploy-daemon has some reducen permissions on the cluster. 

Also, this change worries us a little bit because if the crd got updated with something that is not compatible with some K8s version, we would start having problems in some of our clusters, having to freeze the fiaas-deploy-daemon in those clusters

herodes1991:
Also, I am seeing that Skipper is also in charge of creating/updating the CRD. 
Could this becomes a problem if skipper overwrites the CRD while fiaas-deploy-daemon is running?

herodes1991:
I got it. 
As a summary: 

> this change is intended for two reasons: 
> 1. We want to rid of skipper eventually.
> 2. This is the easier way to support v1beta1 and v1 at the same time

Is this right?

oyvindio:
> 1. We want to rid of skipper eventually.

Yes we do want to get rid of skipper eventually (as I gather this was the general consensus in #163)

> 2. This is the easier way to support v1beta1 and v1 at the same time

Yes, at least it was the best solution we could come up with. 
[...]

PR #175, opened 2022-03-09
Add deactivate-crd-creation config flag

herodes1991:
The new behaviour of the watcher is breaking the fiaas-deploy-daemon in some namespaces of our current architecture. 

We have not the permissions of creating crd in all the namespaces, meaning that fiaas-deploy-daemon crashes and it is not able to start.

This is an implementation to fix the issue #174

### DIFF
--- a/docs/operator_guide.md
+++ b/docs/operator_guide.md
@@ -187,6 +187,10 @@ As we decided to treat the 404 as a valid response, we have no way to differenti
 
 Used to create a serviceaccount for each deployed application, using the application name. If there are imagePullSecrets set on the 'default' service account, these are propagated the per-application service accounts. If a service account with the same name as the application already exists, the application will run under that service account but FIAAS will not overwrite/manage the service account.
 
+### deactivate-crd-creation
+
+Used to deactivate the crd creation in the Crd Watcher.
+
 Deploying an application
 ------------------------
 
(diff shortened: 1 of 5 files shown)


fiaas_deploy_daemon/config.py
```diff
@@ -228,6 +228,9 @@ def _parse_args(self, args):
         parser.add_argument("--use-apiextensionsv1-crd",
                             help="Use apiextensions/v1 CustomResourceDefinition instead of apiextensions/v1beta1",
                             action="store_true", default=False)
+        parser.add_argument("--deactivate-crd-creation",
```
oyvindio, line 231: Nitpicking, but I think naming the flag `disable-crd-creation` would make it a bit more similar to some of the other config flags compared to `deactivate-crd-creation`, since there are a lot of `enable-$feature` flags

docs/operator_guide.md
```diff
@@ -187,6 +187,10 @@ As we decided to treat the 404 as a valid response, we have no way to differenti
 
 Used to create a serviceaccount for each deployed application, using the application name. If there are imagePullSecrets set on the 'default' service account, these are propagated the per-application service accounts. If a service account with the same name as the application already exists, the application will run under that service account but FIAAS will not overwrite/manage the service account.
 
+### deactivate-crd-creation
+
+Used to deactivate the crd creation in the Crd Watcher.
```
oyvindio, line 192: I think some more details about how this flag might be used could be helpful here. How about something like:

"By default fiaas-deploy-daemon updates the Application and ApplicationStatus CRDs on startup. CRDs are cluster scoped, so it is only necessary for one fiaas-deploy-daemon instance in a cluster to manage the CRDs. This flag can be used to disable updating of the CRDs per fiaas-deploy-daemon instance when not needed.
It is a good idea to keep the flag enabled on at least one fiaas-deploy-daemon instance in a cluster, otherwise you have to manually ensure that the Application and ApplicationStatus CRDs are present and up to date in the cluster."

oyvindio (commented):
I think the implementation here looks good. I've added some suggestions about the documentation for and naming of the flag itself inline

oyvindio (approved):
LGTM

PR #176, opened 2022-03-17
Add initial roadmap

oyvindio:
This is a suggestion for a roadmap consisting of some items we've discussed internally. Consider this not written in stone; I'm adding it here for transparency's sake, and as a tool for further discussion.

### DIFF
--- a/docs/roadmap.md
+++ b/docs/roadmap.md
@@ -1,0 +1,23 @@
+# Roadmap
+
+This is a non-exhaustive list of changes we might want to make, in tentatively prioritized order at the time of writing.
+
+T-shirt size: S, M, L indicates a rough estimate of the relative amount of work involved.
+
+- [S] Implement new release model for fiaas-deploy-daemon (#163)
+  - Releases following semantic versioning for fiaas-deploy-daemon
+  - Deprecate fiaas/skipper
+- [L] Run fiaas-deploy-daemon on Python 3 (#6)
+- [S] Make number of ApplicationStatus resources to keep configurable (#46)
+- [M] Deprecate/remove fiaas.yml v2 format
+- [S] Evaluate container registries for moving container images out of DockerHub
+- [M] Support autoscaling on custom metrics for applications
+- [M] Support managing PodDisruptionBudgets for applications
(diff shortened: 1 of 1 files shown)


xamebax (approved):
I wholeheartedly approve this message.

docs/roadmap.md
```diff
@@ -0,0 +1,23 @@
...
+- [S] Implement new release model for fiaas-deploy-daemon (#163)
+  - Releases following semantic versioning for fiaas-deploy-daemon
+  - Deprecate fiaas/skipper
+- [L] Run fiaas-deploy-daemon on Python 3 (#6)
+- [S] Make number of ApplicationStatus resources to keep configurable (#46)
+- [M] Deprecate/remove fiaas.yml v2 format
+- [S] Evaluate container registries for moving container images out of DockerHub
+- [M] Support autoscaling on custom metrics for applications
+- [M] Support managing PodDisruptionBudgets for applications
+- [L] Look into replacing fiaas/k8s with kubernetes-clients/python
```
mortenlj, line 16: I'd like to hear a bit of background for this item?

Last time I checked, the official client was really cumbersome to work with, especially when dealing with CRDs.
If it was up to me, I would probably prefer to put my time into improving fiaas/k8s in the places where it falls short rather than trying to work with the official client. The only thing I really would like to get from the official client is the config-loading mechanisms.

xamebax: The background is partially captured by the list of PRs closed in the k8s repository, for example this one: https://github.com/fiaas/k8s/pull/101

We shouldn't be spending time maintaining our own client, because each time an API changes, we have to reflect that. This increases our workload when having to make the FIAAS ecosystem compatible with a new Kubernetes version, and as you know, these come regularly.

oyvindio: > We shouldn't be spending time maintaining our own client, because each time an API changes, we have to reflect that. This increases our workload when having to make the FIAAS ecosystem compatible with a new Kubernetes version, and as you know, these come regularly.

I think this is a good summary of the background/motivation for this. 
Also I see this as initially an exploratory task, to determine if using the official client could be a good fit, before either committing to switching or not.

PR #177, opened 2022-03-23
add documentation on annotations/labels for serviceaccounts

j18e:
This adds service_accounts to the list of resource types which may be annotated/labeled via fiaas.yml in the v3 spec documentation.


tg90nor (approved):
LGTM

Issue #178, opened 2022-03-31
Change the way extra ingresses are created


herodes1991:
These days we are migrating some ingresses to migrate them from HTTP01 to DNS and we are facing some problems.

Due to this migrations, some requests that were previously working, are starting to respond with error while the certificate is not validated. Our idea is to copy the secrets from the original ingress to the new ones to avoid this downtime.

oyvindio:
Is it possible to add some more details to describe this issue? It is not entirely clear to me which situation or configuration(s) that make this a problem and because of that it is difficult for me to consider the benefit/potential drawbacks of the suggested solution

herodes1991:
Yes! For example when you have the ingress with tls enabled defined as internal and you want to create an extra ingress defined as nginx. During the time where the certificate is being created, the nginx one will not be working. With this change the nginx one will use the same certificate as the internal one until the new certificate is created.

oyvindio:
I'm sorry, I still don't entirely understand this. I'd appreciate if you could help make some of these points clearer to me:

> For example, when you have the ingress with tls enabled defined as internal and you want to create an extra ingress defined as nginx.

- What do you mean by "defined as internal" and "defined as nginx" here? Does it have to do with certificate issuers or ingress class or something else entirely?
- How do the two ingresses look (hosts, paths, relevant annotations etc.) in this example?

In the description of the issue you mentioned this:

> These days we are migrating some ingresses to migrate them from HTTP01 to DNS and we are facing some problems.

Does this refer to a "one-off" migration to switch [cert-manager ACME solvers](https://cert-manager.io/docs/configuration/acme/#solving-challenges) for some/all ingresses?

portega-adv:
Let me try to help understanding the issue:

About HTTP/DNS migration: yes, it refers to cert manager and cert challenges. While HTTP has no delay (once you publish the challenge response file, it's available), moving to DNS challenge adds a delay for DNS propagation. Copying secrets here helps to avoid that, by providing the current certificate until the new one is generated. 

This will be a one time operation, but there are other cases where the change will be helpful too, and those will be recurring cases. The internal/nginx example refers to that, when you add annotations and the update will enforce the Ingress to split in two objects, one per domain, the "whatever-app" Ingress will keep the certificate but the "whatever-app-1" will not, and will be unavailable until a new one will be generated.

[...]

PR #180, opened 2022-05-31
Implement semantic versioning based release process

### LINKED ISSUE #179: Disable arm container image builds
I've started looking into setting up the new release model for fiaas-deploy-daemon (#163). I think it makes sense to do that in a similar manner to fiaas/k8s, where there is a [separate build stage](https://github.com/fiaas/k8s/blob/master/.semaphore/semaphore.yml#L51-L57) for creating the release, which is only run for builds triggered by pushing a git tag. I'd also like to set up the e2e tests to run in parallel using a matrix build, since that should reduce CI build duration, and would speed up testing of the release builds.

I think it is important to use the same container image throughout the CI pipeline; from initial build stage, to the e2e build stages, to the release build stage. Unfortunately this seems to me to be a bit tricky to do with the way the fiaas-deploy-daemon container image is currently built. The current CI build builds a single-manifest multi-arch image for amd64, arm64 and arm7. This type of image can be exported as a tarball with `docker buildx --output=type=tar`, which the current build does, but it doesn't seem to be possible to actually import this tarball into docker (in later build stages), so that can be used to run a container or be pushed to a container registry. At least I haven't been able to figure out how. There also seems to be a few tickets on buildx related to this capability:
- https://github.com/docker/buildx/issues/847
- https://github.com/docker/buildx/issues/430
- https://github.com/docker/buildx/issues/393

Simple "single arch" images built `docker build` (i.e. not buildx) can be exported and imported to/from file with `docker image save/load`, and be moved through a CI pipeline that way. Building arm images [was disabled in fiaas/skipper](https://github.com/fiaas/skipper/pull/109) almost a year ago. That makes disabling arm container image builds for fiaas-deploy-daemon to work around the apparent limitations in the tooling for multi-arch images seem like a potential option, unless someone actually depends on the arm container images.

If you are using fiaas-deploy-daemon container images built for arm64/arm7, please comment here and we can discuss alternative solutions.

oyvindio:
This PR implements the release model described in #163.

Notable changes:
- Add a helm chart for deploying fiaas-deploy-daemon
- `latest` and `stable` release channels for use with skipper (in [fiaas/releases](https://github.com/fiaas/releases)) *will no longer be updated*
- CI builds of commits on the master branch will push a versioned container image `fiaas/fiaas-deploy-daemon:$timestamp-$commit_ref` like before, and will also update the `fiaas/fiaas-deploy-daemon:development` container image tag to point to the same image ref. The `:latest` container image tag will no longer be updated (it looks like this tag haven't actually been pushed for some time).
- CI builds of annotated tags matching `vMAJOR.MINOR.PATCH` (e.g. `v1.2.3`) will 
  - push a versioned container image `fiaas/fiaas-deploy-daemon:$timestamp-$commit_ref` as well as a tag for the release `fiaas/fiaas-deploy-daemon:vMAJOR.MINOR.PATCH`. 
  - update the helm chart with the container image reference and chart version to match the release version, package it and push it to [fiaas/helm](https://github.com/fiaas/helm).
  - create a github release containing release notes from the annotated tag
- The e2e tests run in a matrix build (in parallell) in CI, parameterized by target Kubernetes version. This reduces the duration of the CI build significantly.
  - A Kubernetes version parameter is added to the e2e test. When invoked locally, e2e tests only run against the latest configured Kubernetes version; this is a change from the previous behavior of running the tests against all configured Kubernetes versions. To run the e2e tests against a specific Kubernetes version, run `tox -e integration_test -- --e2e-k8s-version $k8s_version`
  - The list of target Kubernetes versions is moved  to [.semaphore/semaphore.yaml](https://github.com/fiaas/fiaas-deploy-daemon/blob/163-semver-releases/.semaphore/semaphore.yml#L62-L70). The default Kubernetes version used when running `tox -e integration_test` is still configured in [tests/conftest.py](https://github.com/fiaas/fiaas-deploy-daemon/blob/163-semver-releases/tests/conftest.py#L160-L162)
- When running e2e tests with `--use-docker-for-e2e`, the test now uses the `:development` (and not `:latest`) container image tag for fiaas/fiaas-deploy-daemon.
- The fiaas-deploy-daemon container image is only built for amd64. Multi-arch builds for arm64 and arm7 are disabled (see #179)

I've updated some of the documentation for how CI works and documented how to create a release. I've also added some documentation for how to migrate from skipper to the fiaas-deploy-daemon helm chart, but as there are a lot of details and configurations can vary please consider that a high level guideline and not a complete step-by-step recipe.

I'm going to create some test releases to verify that the CI promotion step for releases works before merging this by tagging some commits on this branch, as I'm not sure how to test that without actually creating a release. This should also make it possible to verify that the helm chart works as expected in different configurations before creating the first proper release.

Closes #179


oyvindio:
I've created a test release [v0.0.1-alpha.2](https://github.com/fiaas/fiaas-deploy-daemon/releases/tag/v0.0.1-alpha.2), and that seemed to work as expected. I tried installing fiaas-deploy-daemon via the published helm chart (`helm install --upgrade fiaas-deploy-daemon fiaas/fiaas-deploy-daemon`), and it only seems to work when `--version v0.0.1-alpha.2` is specified explicitly (otherwise it fails with `Error: failed to download "fiaas/fiaas-deploy-daemon" (hint: running 'helm repo update' may help)`. I guess that might be because `v0.0.1-alpha.2` is a pre-release version as defined by semver and helm doesn't recognize that version as the latest release.
In any case, when specifying the version explicitly install seems to work with both helm2 and helm3.

helm/fiaas-deploy-daemon/templates/deployment.yaml
```diff
@@ -0,0 +1,87 @@
...
+    {{- include "fiaas-deploy-daemon.labelsOrAnnotations" .Values.annotations.global | indent 4 }}
+    {{- include "fiaas-deploy-daemon.labelsOrAnnotations" .Values.deployment.annotations | indent 4 }}
+{{- end }}
+spec:
+  replicas: 1
+  strategy:
+    rollingUpdate:
+      maxSurge: 0
+      maxUnavailable: 1
+    type: RollingUpdate
```
j18e, line 19: What's the advantage of using RollingUpdate over Recreate as the strategy?

j18e (commented):
Looks really good, there's a lot of high quality work in here! I have one quick question but other than that it looks good.

helm/fiaas-deploy-daemon/templates/deployment.yaml
oyvindio: There is no technical reason: I initially used Recreate when creating the chart, but changed it to RollingUpdate only because that's how fiaas-deploy-daemon deploys itself now, and I thought that keeping it like that would be one less change in this PR. It is perhaps more "correct" to use Recreate, since having more than one fiaas-deploy-daemon running in a namespace is not a good idea as that could for example cause duplicate deploys. What do you think?

j18e: I think it would be a small improvement in the chart, and i can't see how it would negatively affect anything to change it to Recreate here.

portega-adv:
First of all, amazing job doing this MASSIVE update, kudos for that!!

I know for sure we'll need a bit of time to review it, but anyway, as you asked for concerns too, let me share one: We are working in a small fix for issue #178 and I hope we'll have time to apply it before moving into semver. So I'll ask you to wait (if necessary) until this change is merged.

oyvindio:
> First of all, amazing job doing this MASSIVE update, kudos for that!!

Thanks for the feedback 👍 

> I know for sure we'll need a bit of time to review it, but anyway, as you asked for concerns too, let me share one: We are working in a small fix for issue https://github.com/fiaas/fiaas-deploy-daemon/issues/178 and I hope we'll have time to apply it before moving into semver. So I'll ask you to wait (if necessary) until this change is merged.

I think we can try to accommodate that and wait with merging this until #178 is resolved

bin/make_release
```diff
@@ -0,0 +1,121 @@
...
+
+function publish_helm_chart() {
+    local release="$1"
+
+    update_chart_release_version "$release"
+    # helm package outputs 'Successfully packaged chart and saved it to: path/to/chart.tgz',
+    # remove 'Successfully packaged ...' prefix to get path to chart tarball
+    local helm_package_output
+    helm_package_output="$(helm package helm/fiaas-deploy-daemon)"
+    local chart_tarball="${helm_package_output##Successfully packaged chart and saved it to: }"
```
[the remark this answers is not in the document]
oyvindio: The reason for parsing the output of the command here is to get the full path to the tarball, including its filename. The `-d` option only makes it possible to set the directory the tarball is created in, so to get the full path from that it would be necessary to guess or infer the filename.

oyvindio:
I have just pushed some minor changes to use the `Recreate` deployment strategy in the chart and fix a typo in the documentation, and I've created another pre-release tag just to have a release of the tip of the branch for testing.

[Tag protection](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/configuring-tag-protection-rules) is now enabled on this repo to limit who can push tags and create releases. Pushing a tag should require admin privileges at the repo level, which members of the maintainer team should have. Tag protection is apparently a beta feature, so if it happens to cause any problems later it can be changed in the repo settings.

oyvindio:
This PR has been open for a while now with few significant changes, so I would like to merge it early next week to move forward with the release model changes. @fiaas/maintainers please take a look and let me know if you have any feedback or concerns. If there are no objections I will go ahead and merge this on Tuesday next week (2022-09-13).

j18e (approved):
LGTM!

PR #181, opened 2022-06-08
Copy tls ingress secrets

portega-adv:
Fix for issue #178 

As we're splitting and creating new Ingress objects, the certificates may need to be renewed, so there's a time gap while we wait for the new certificate, that the app will not be available. We can avoid this delay copying the certificate Secret to the new Ingress.


fiaas_deploy_daemon/deployer/kubernetes/ingress.py
```diff
@@ -48,18 +51,61 @@ def delete(self, app_spec):
         LOG.info("Deleting ingresses for %s", app_spec.name)
         self._ingress_adapter.delete_list(app_spec)
 
+    def _list_secrets(self, app_spec):
```
xavileon, line 54: nit: _map_hostnames_to_certificate_secrets(self, app_spec)

fiaas_deploy_daemon/deployer/kubernetes/ingress.py
```diff
@@ -48,18 +51,61 @@ def delete(self, app_spec):
...
+        hosts_map = {}
+        for k8s_ingress in k8s_ingresses:
+            for tls_obj in k8s_ingress.spec.tls:
+                if tls_obj.secretName:
+                    for host_item in tls_obj.hosts:
+                        hostname = host_item.strip()
+                        hosts_map[hostname] = tls_obj.secretName
+        return hosts_map
+
+    def _create_secret(self, secret_name, new_name, app_spec):
```
xavileon, line 66: nit: _copy_secret(self, old_secret_name, new_secret_name, app_spec)

fiaas_deploy_daemon/deployer/kubernetes/ingress.py
```diff
@@ -48,18 +51,61 @@ def delete(self, app_spec):
...
+        hosts_map = self._list_secrets(app_spec)
+
+        LOG.info("Will create %d ingresses", len(ingresses))
         for annotated_ingress in ingresses:
             if len(annotated_ingress.ingress_items) == 0:
                 LOG.info("No items, skipping: %s", annotated_ingress)
                 continue
 
+            if len(hosts_map) > 0:
+                new_name = "{}-ingress-tls".format(annotated_ingress.name)
```
xavileon, line 96: I'd extract this format string to a separate function, as it's used in the [IngressTLSDeployer](https://github.com/fiaas/fiaas-deploy-daemon/pull/181/files#diff-b57272c01d7e5fbc0b4cfcbe9654ffde84de2a45e6bf3f1a5231eb11f478d036R261) to define the name of the secret.

fiaas_deploy_daemon/deployer/kubernetes/ingress.py
```diff
@@ -48,18 +51,61 @@ def delete(self, app_spec):
...
+                try:
+                    found_secret = Secret.get(new_name, app_spec.namespace)
+                except NotFound:
+                    found_secret = None
+
+                if not found_secret:
+                    for ingress_item in annotated_ingress.ingress_items:
+                        secret_name = hosts_map[ingress_item.host]
```
xavileon, line 104: nit: if you use `hosts_map.get(ingress_item.host)` then you don't need to check the size of the map in line 95.

xavileon (commented):
a few nits for readibility. General question: should the new code belong the the `IngressTLSDeployer` object given that copying secrets will inly happen if TLS is enabled?

fiaas_deploy_daemon/deployer/kubernetes/ingress.py
portega-adv: The idea is to skip the whole section if there's no map to check (just a little performance improvement)

bin/run_fdd_against_kind
```diff
@@ -33,6 +33,7 @@ ARGS=(
 '--datadog-container-image' 'datadog/docker-dd-agent:12.2.5172-alpine'
 '--enable-crd-support'
 '--enable-service-account-per-app'
+'--use-ingress-tls' 'default_off'
```
herodes1991, line 36: Should we reverse this change?

herodes1991 (approved):

tests/fiaas_deploy_daemon/test_e2e.py
```diff
@@ -549,6 +550,104 @@ def cleanup_complete():
...
+
+            for ingress_name, expected_dict in expected.items():
+                actual = k8s_ingress.get(ingress_name)
+                assert_k8s_resource_matches(actual, expected_dict, IMAGE1, None, DEPLOYMENT_ID2, None, app_uid)
+
+        def _check_two_secrets():
+            assert Secret.get("{}-ingress-tls".format(name))
+            assert Secret.get("{}-1-ingress-tls".format(name))
+
+        def _assert_status2():
```
xavileon, line 631: nit: _assert_status_after_splitting_ingress?

fiaas_deploy_daemon/deployer/kubernetes/ingress.py
```diff
@@ -48,18 +51,62 @@ def delete(self, app_spec):
...
+        if old_secret:
+            new_metadata = ObjectMeta(
+                annotations=old_secret.metadata.annotations,
+                labels=old_secret.metadata.labels,
+                name=new_name,
+                namespace=old_secret.metadata.namespace
+            )
+            new_secret = Secret(metadata=new_metadata, data=old_secret.data, type=old_secret.type)
+            new_secret.save()
+
```
xavileon, line 80: finding which secret to copy and copying the secret seem to belong to the IngressTLSDeployer in terms of responsability instead of the main IngressDeployer. We will need to get all ingresses for each ingress we create, but that's a performance penalty worth considering in exchange for code clarity (e.g. one extra "find" call for each ingress, and the common case is to have just one ingress in the app). What are your thoughts on this?

xavileon (commented):
good job! I added a few minor comments and a question about where the logic of this change should land wrt to code responsbility.

bin/run_fdd_against_kind
portega-adv: Right, is not needed in the general use case

fiaas_deploy_daemon/deployer/kubernetes/ingress.py
```diff
@@ -48,18 +51,62 @@ def delete(self, app_spec):
...
+                try:
+                    found_secret = Secret.get(new_name, app_spec.namespace)
+                except NotFound:
+                    found_secret = None
+
+                if not found_secret:
+                    for ingress_item in annotated_ingress.ingress_items:
+                        secret_name = hosts_map[ingress_item.host]
+                        if secret_name:
+                            LOG.info("Copying secrets")
```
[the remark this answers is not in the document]
portega-adv: Sure, a lot more useful!

tests/fiaas_deploy_daemon/test_e2e.py
portega-adv: Aboslutely

fiaas_deploy_daemon/deployer/kubernetes/ingress.py
portega-adv: After a bit of thinking, we agreed to move the copy secrets code to the TLSDeployer, but keeping the hosts map function in place, to improve efficiency. WDYT?

herodes1991 (approved):

portega-adv:
We decided to close the issue and use a different approach

PR #182, opened 2022-07-12
Fix bug with ingress names

herodes1991:
During these days we discovered a bug with the ingresses creation.

When you have an ingress with a host configured with a different issue type than default, and then another one with annotations, the ingress name of both ingresses will be the same, as the `ingresses` variable will not be modified by the `override_issuer_ingresses` dictionary ones, being these ones added at the end.


xavileon (approved):
🔝

PR #183, opened 2022-07-14
Added check for Ingress certificates readiness

portega-adv:
We keep working around the issues we have with how Ingress objects are managed (see #178 and #182 ).

In this case, we want to check if the TLS certificates are ready before we start to send traffic to the related Ingress, avoiding connection errors.

This PR will add a new parameter to check certificates status (default being False to keep backwards compatibility)

### DIFF
--- a/fiaas_deploy_daemon/config.py
+++ b/fiaas_deploy_daemon/config.py
@@ -278,6 +278,7 @@ class Configuration(Namespace):
         tls_parser.add_argument("--tls-certificate-issuer-type-overrides", help="Issuers to use for specified domain suffixes",
                                 default=[],
                                 action="append", type=KeyValue, dest="tls_certificate_issuer_type_overrides")
+        tls_parser.add_argument("--tls-certificate-ready", help="Check whether certificates are ready", default=False)
 
         parser.parse_args(args, namespace=self)
         self.global_env = {env_var.key: env_var.value for env_var in self.global_env}
--- a/fiaas_deploy_daemon/config.py
+++ b/fiaas_deploy_daemon/config.py
@@ -278,7 +278,7 @@ class Configuration(Namespace):
         tls_parser.add_argument("--tls-certificate-issuer-type-overrides", help="Issuers to use for specified domain suffixes",
                                 default=[],
                                 action="append", type=KeyValue, dest="tls_certificate_issuer_type_overrides")
-        tls_parser.add_argument("--tls-certificate-ready", help="Check whether certificates are ready", default=False)
+        tls_parser.add_argument("--tls-certificate-ready", help="Check whether certificates are ready", default=False, action="store_true")
 
         parser.parse_args(args, namespace=self)
         self.global_env = {env_var.key: env_var.value for env_var in self.global_env}
--- a/fiaas_deploy_daemon/deployer/kubernetes/ready_check.py
+++ b/fiaas_deploy_daemon/deployer/kubernetes/ready_check.py
@@ -66,8 +66,7 @@ class ReadyCheck(object):
                 dep.status.observedGeneration >= dep.metadata.generation)
 
     def _is_certificate_ready(self, cert):
-        conditions = cert.status.conditions
-        for condition in conditions:
+        for condition in cert.status.conditions:
             if condition.type == "Ready":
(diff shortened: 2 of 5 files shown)


fiaas_deploy_daemon/config.py
```diff
@@ -278,6 +278,7 @@ def _parse_args(self, args):
         tls_parser.add_argument("--tls-certificate-issuer-type-overrides", help="Issuers to use for specified domain suffixes",
                                 default=[],
                                 action="append", type=KeyValue, dest="tls_certificate_issuer_type_overrides")
+        tls_parser.add_argument("--tls-certificate-ready", help="Check whether certificates are ready", default=False)
```
herodes1991, line 281: Instead of allowing any kind of value, add an `action="store_true"` like the options of enable/disable (for example `disable-deprecated-managed-env-vars`)

fiaas_deploy_daemon/deployer/kubernetes/ready_check.py
```diff
@@ -62,6 +65,35 @@ def _ready(self):
...
+                return condition.status == "True"
+        return False
+
+    def _ingress_ready(self):
+        try:
+            ing_list = Ingress.find(self._app_spec.name, self._app_spec.namespace)
+        except NotFound:
+            return False
+
+        for ingress in ing_list:
```
herodes1991, line 81: You should check if the ingress has tls, before checking the certificate

tests/fiaas_deploy_daemon/deployer/kubernetes/test_ready_check.py
```diff
@@ -114,6 +116,39 @@ def test_deployment_complete_deactivated(self, get, app_spec, bookkeeper, lifecy
...
+        replicas = 2
+
+        with mock.patch("k8s.models.ingress.Ingress.find") as find:
+            find.return_value = []
+
+            self._create_response(get, replicas, replicas, replicas, replicas)
+            ready = ReadyCheck(app_spec, bookkeeper, lifecycle, lifecycle_subject, config)
+
+            assert ready() is False
+
```
herodes1991, line 151: Add test to ensure that, for the case of ingress without tls, the method returns true

herodes1991 (changes requested):

herodes1991 (approved):
LGTM, great job! 🏆

fiaas_deploy_daemon/deployer/kubernetes/ready_check.py
```diff
@@ -62,6 +65,36 @@ def _ready(self):
                 dep.status.availableReplicas == expected_value and
                 dep.status.observedGeneration >= dep.metadata.generation)
 
+    def _is_certificate_ready(self, cert):
```
fpiecho, line 68: Should we add the `notAfter` condition here?

portega-adv: Sure thing! Thanks for the reminder

fiaas_deploy_daemon/deployer/kubernetes/ready_check.py
```diff
@@ -66,6 +68,8 @@ def _deployment_ready(self):
                 dep.status.observedGeneration >= dep.metadata.generation)
 
     def _is_certificate_ready(self, cert):
+        if cert.status.NotAfter < datetime.utcnow():
```
fpiecho, line 71: Will this fail if `notAfter` is null? I think it can be null according to [docs](https://cert-manager.io/docs/reference/api-docs/)
I think we should also return `false` if `notAfter` is null

portega-adv: notAfter is optional. That means it could be null, and that could mean that it doesn't expire, so I'm afraid failing with null could lead to false negatives...

portega-adv: Also note the default value will be None, and it doesn't fail when it's not set (not with an error, it just takes it as false)

portega-adv: I added a specific test for this case

herodes1991 (approved):

PR #186, opened 2022-09-13
Use port number from values in ingress

oyvindio:
This fixes an issue in some ingress controllers, and also matches what fiaas-deploy-daemon itself does so that the ingress should look the same as when fiaas-deploy-daemon is deployed with skipper/by itself (same as before).


j18e (approved):
Looks good!

PR #191, opened 2022-10-27
Fix versions of ingresses in ready_check method

herodes1991:
During the deployment in newer versions of K8s we found out that the ingress version we were retrieving was the v1beta1. We changed the code to use ingress_adapter and get the correct version depending on the fiaas configuration.

### DIFF
--- a/tests/fiaas_deploy_daemon/deployer/kubernetes/test_ready_check.py
+++ b/tests/fiaas_deploy_daemon/deployer/kubernetes/test_ready_check.py
@@ -20,13 +20,19 @@ import mock
 import pytest
 from monotonic import monotonic as time_monotonic
 
+from fiaas_deploy_daemon import ExtensionHookCaller
 from fiaas_deploy_daemon.deployer.bookkeeper import Bookkeeper
+from fiaas_deploy_daemon.deployer.kubernetes.ingress import IngressTLSDeployer
+from fiaas_deploy_daemon.deployer.kubernetes.ingress_networkingv1 import NetworkingV1IngressAdapter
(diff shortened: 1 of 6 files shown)


tests/fiaas_deploy_daemon/deployer/kubernetes/test_ready_check.py
```diff
@@ -232,10 +265,141 @@ def test_tls_ingress_not_ready(self, get, app_spec, bookkeeper, lifecycle, lifec
...
+        self._create_response(get)
+        ready = ReadyCheck(app_spec, bookkeeper, lifecycle, lifecycle_subject, ingress_adapter, config)
+
+        assert ready() is False
+        bookkeeper.success.assert_called_with(app_spec)
+        bookkeeper.failed.assert_not_called()
+        lifecycle.success.assert_called_with(lifecycle_subject)
+        lifecycle.failed.assert_not_called()
+
+    def test_tls_ingress_ready_v1(self, get, app_spec, bookkeeper, lifecycle, lifecycle_subject, ingress_v1_adapter,
```
portega-adv, line 279: It would be nicer to parametrize the tests so you can test both versions with the same method, and avoid boilerplate

portega-adv (approved):

oyvindio (commented):
I've left some suggestions as inline comments which I think might help simplify the tests a bit. Other than that I think this looks good.

herodes1991:
Hello oyvindio! Can you take a look? We would like to deploy it during next week. What do you think?

oyvindio (approved):
Looks good to me

portega-adv (approved):

PR #192, opened 2022-11-01
Add Kubernetes 1.23 to e2e test Kubernetes version CI targets

oyvindio:
Also set the default target for e2e tests to the same version.


herodes1991 (approved):

bjornakr (approved):
Good to have that dependency documented

PR #194, opened 2022-11-15
Remove older kubernetes versions in e2e test configuration

bjornakr:
We are running e2e-tests on a lot of older versions of Kubernetes. To avoid spending unneccessary resources on the build step, I have removed most of them. I kept the three last entries, just in case.

If anyone is using an older version, please comment here, so I can put it back in.

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -63,11 +63,6 @@ blocks:
           # E2E_K8S_VERSION_OPTION in tests/conftest.py; it should point to the most recent version used in CI.
           - env_var: K8S_VERSION
             values:
-              - v1.15.12
-              - v1.16.15
-              - v1.18.19
-              - v1.19.11
-              - v1.20.7
               - v1.21.2
               - v1.22.4
               - v1.23.13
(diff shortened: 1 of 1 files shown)


.semaphore/semaphore.yml
```diff
@@ -63,11 +63,6 @@ blocks:
           # E2E_K8S_VERSION_OPTION in tests/conftest.py; it should point to the most recent version used in CI.
           - env_var: K8S_VERSION
             values:
-              - v1.15.12
```
herodes1991, line 66: Can we continue testing the `v1.15.12`? We still have this version in our K8s environment 😄

herodes1991: We can remove it once we move out of these clusters 😄

bjornakr: I'll put it back in. 👍 Please let us know when when you move on.

herodes1991 (approved):

bjornakr:
@birgirst Does this look ok to you?

bjornakr:
@PerGon Could you check if this is ok for developer foundations?

birgirst (approved):
Lgtm

Issue #195, opened 2022-11-15
Problem with application status


herodes1991:
We just detected a problem with the application status result. We have a user that set a smaller readiness timeout than the liveness and the app is constantly finishing in a FAILED status. What do you think about getting the grater value between readiness and liveness?

https://github.com/fiaas/fiaas-deploy-daemon/blob/727967d9543d6fc0d6f9284421461a0e131a2f1b/fiaas_deploy_daemon/deployer/kubernetes/ready_check.py#L39

oyvindio:
> We just detected a problem with the application status result. We have a user that set a smaller readiness timeout than the liveness and the app is constantly finishing in a FAILED status.

I think some more details might be necessary to understand what the problem is in this case. Does the deployment rollout complete successfully at the Kubernetes level? How are the healthchecks for this application configured? Is it possible to create an example application configuration/application resource which can reproduce the issue?

If the deployment rollout does not complete successfully, the healthchecks/readiness probe for the application might need to be adjusted. If the rollout completes successfully but takes longer than the ReadyCheck timeout because of external factors such as image pull or pod scheduling delays, one option could be to increase the `ready-check-timeout-multiplier` config flag. Note that this will increase the effective ReadyCheck timeout for all applications deployed by the fiaas-deploy-daemon instance.

> What do you think about getting the grater value between readiness and liveness?

ReadyCheck uses `initial_delay_seconds` from the readiness probe because during a deployment rollout it is necessary to wait for the readiness probes for each pod to go to `Success` to ensure that an appropriate number of pods are available at all times. My understanding is that since the default state of liveness probes is `Success`, the deployment controller does not wait for the liveness probe `initial_delay_seconds` during rollout. Liveness probes can only influence the result of a rollout by failing.
ReadyCheck tries to determine whether the deployment rollout was successful within a timeout based on an estimate of how long a rollout might take to complete. If liveness probe `initial_delay_seconds` isn't a factor in how much time rollouts take in practice, then I'm not sure that using it to calculate the ReadyCheck timeout is an ideal solution.

herodes1991:
The problem is that the app is not even healthy. It's true that they need to improve the apps creation, but it's weird that fiaas application status becomes "FAILED" before giving the pods the option to become healthy

herodes1991:
oh, yes, it was a typo, it was 120 seconds 😅

oyvindio:
[...]

PR #196, opened 2022-11-15
Use maximum between liveness and readiness for fail_time

herodes1991:
https://github.com/fiaas/fiaas-deploy-daemon/issues/195

### DIFF
--- a/fiaas_deploy_daemon/deployer/kubernetes/ready_check.py
+++ b/fiaas_deploy_daemon/deployer/kubernetes/ready_check.py
@@ -33,10 +33,12 @@ class ReadyCheck(object):
         self._bookkeeper = bookkeeper
         self._lifecycle = lifecycle
         self._lifecycle_subject = lifecycle_subject
+        readiness_delay = app_spec.health_checks.readiness.initial_delay_seconds
+        liveness_delay = app_spec.health_checks.liveness.initial_delay_seconds
         self._fail_after_seconds = _calculate_fail_time(
             config.ready_check_timeout_multiplier,
             app_spec.autoscaler.max_replicas,
-            app_spec.health_checks.readiness.initial_delay_seconds
+            readiness_delay if readiness_delay > liveness_delay else liveness_delay
         )
         self._fail_after = time_monotonic() + self._fail_after_seconds
         self._should_check_ingress = config.tls_certificate_ready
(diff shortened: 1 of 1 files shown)


fiaas_deploy_daemon/deployer/kubernetes/ready_check.py
```diff
@@ -33,10 +33,12 @@ def __init__(self, app_spec, bookkeeper, lifecycle, lifecycle_subject, ingress_a
         self._bookkeeper = bookkeeper
         self._lifecycle = lifecycle
         self._lifecycle_subject = lifecycle_subject
+        readiness_delay = app_spec.health_checks.readiness.initial_delay_seconds
+        liveness_delay = app_spec.health_checks.liveness.initial_delay_seconds
         self._fail_after_seconds = _calculate_fail_time(
             config.ready_check_timeout_multiplier,
             app_spec.autoscaler.max_replicas,
-            app_spec.health_checks.readiness.initial_delay_seconds
+            readiness_delay if readiness_delay > liveness_delay else liveness_delay
```
portega-adv, line 41: I'd prefer to use max(readiness_delay, liveness_delay) instead, clean and easy to read.

oyvindio:
It is possible that I don't have all the details, but based on my current understanding of the issue in #195 , it seems to me like the changes in this PR as it currently is could have some potential drawbacks:
ReadyCheck tries to set a timeout based on the expected duration of a successful rollout at the Kubernetes level. During a rollout it is necessary to wait for initial_delay_seconds from the readiness probe for each pod to be ready, to ensure there are always some number of available pods. As far as I understand, the rollout process does not wait for initial_delay_seconds set on the liveness probe, because this probe's default state is success.
This means readiness probe initial_delay_seconds influence rollout duration directly, but the liveness probe initial_delay_seconds setting to my knowledge doesn't. Because of this difference, it seems to me like a ReadyCheck timeout set based on liveness probe initial_delay_seconds would not necessarily be a good estimate of the rollout duration. Even if the value of initial_delay_seconds on the liveness probe is higher than the same parameter on the readiness probe, the way I see it, a longer timeout isn't always better because it means it takes a longer time before the FAILED result is set on the application status if the rollout fails, and the user deploying the application can get feedback about that.

herodes1991:
I think I am agree with you @oyvindio, the only weird thing is that we had the scenario where the deployment was successfully deployed but the FIAAS application was marked as FAILED 😔. I will close it for now and encourage our users to increase the initial_delay to be more "flexible" 😄

PR #197, opened 2022-12-16
Use python 3.9

### LINKED ISSUE #6: Upgrade to python 3
Work is underway to remove integration with Kafka that uses python 2.7 libraries.
Once that is complete fiaas-deploy-daemon can be upgraded to python 3.

oyvindio:
Switch from python 2.7 to 3.9.

I picked python 3.9 to get up to python 3.x with the least amount of changes, as using 3.10 or 3.11 will require updating some dependencies since 3.10 removes a few things that have been deprecated for some time.

Summary of changes:
- Automatically updated all code with `2to3`. This is mostly removal of unicode literals and `__future__` imports, wrapping calls to APIs which have changed to return a different iterable type in `list()` and a few other changes.
- Replace `hash()` with [py27hash](https://pypi.org/project/py27hash/) for generating ApplicationStatus names to ensure the same names as python2 fiaas-deploy-daemon would generate.
- Several fixes for transforming bytes to unicode and vice versa.
- Use `unittest.mock` and `time.monotonic` from stdlib, drop 3rd party dependencies.
- pin Flask related dependencies, upgrading from flask 1.x to 2.x should be done later.
- Fix issue with running e2e tests requiring the same fiaas-deploy-daemon + kind cluster fixures on different worker processes. When this happens additional time is spent starting a kind cluster and fiaas-deploy-daemon combination which is already available in another test worker, which increased test duration. Total e2e test duration should now be a bit faster and more consistent.

closes #6


henrik242 (approved):
Lgtm

birgirst:
Awesome work @oyvindio ! I’ll have a look through.

birgirst (approved):
LGTM

PR #198, opened 2022-12-30
Add status field to FIAAS Application resource

mika-hjalmarsson:
Early PR to get feedback on adding status field to FIAAS Application resource.
This PR will only add the status field and continue to use ApplicationStatus resource as well.

The status field is similar to ApplicationStatus but only logs error, to not bloat the Application resource.

Ref. the roadmap where this feature is highlighted.
https://github.com/fiaas/fiaas-deploy-daemon/blob/master/docs/roadmap.md

### DIFF
--- a/fiaas_deploy_daemon/crd/types.py
+++ b/fiaas_deploy_daemon/crd/types.py
@@ -17,8 +17,10 @@
 from __future__ import absolute_import
 
 import six
-from k8s.base import Model
-from k8s.fields import Field, RequiredField, ListField
+from k8s.base import Model, SelfModel
+from k8s.fields import Field, RequiredField, ListField, JSONField
+from k8s.models.apiextensions_v1_custom_resource_definition import CustomResourceSubresourceScale,\
+    ExternalDocumentation
 from k8s.models.common import ObjectMeta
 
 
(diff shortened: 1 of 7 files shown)


fiaas_deploy_daemon/crd/types.py
```diff
@@ -56,6 +58,24 @@ class Meta:
     spec = Field(FiaasApplicationSpec)
 
 
+class FiaasApplicationStatusResult(Model):
+    result = RequiredField(six.text_type)
+    observedGeneration = Field(int, 0)
+    logs = ListField(six.text_type)
+
+
+class FiaasApplicationStatusInline(Model):
```
mortenlj, line 67: This feels like a hack.

The proper way to do this is probably to implement support for working with subresources of any resource that has those in the k8s library.

fiaas_deploy_daemon/log_extras.py
```diff
@@ -34,6 +35,16 @@ def filter(self, record):
         return 1
 
 
+class ExtraErrorFilter(logging.Filter):
+    def filter(self, record):
+        extras = {}
+        for key in ("app_name", "namespace", "deployment_id"):
```
mortenlj, line 41: It feels wrong to do both filtering and adding extra-values in the same filter.
It would make sense to separate the two and apply both filters, just to separate the concerns.

oyvindio (commented):
Thank you for the PR! Some general comments;

- I think that when implementing this it is important to maintain backwards compatibility, so that the ApplicationStatus resource still behaves as it currently does. It sounds like that is the intention from the PR description, but I thought I would mention it also.

- I could be wrong, but I think that if one deploy (1) is waiting for its ReadyCheck to complete, and another deploy (2) happens on the same application before 1 completes, there could be the possibility of a race condition where deploy 1's ReadyCheck could overwrite the status/result for deploy 2. Based on the level-driven principle of the Kubernetes API, I think the behaviour in that case should be that deploy 2 should take priority and the result of deploy 1's ReadyCheck should be ignored. I'm not sure how that would be best implemented though.

I've left some suggestions/comments in line also

herodes1991:
Will this be continued? 😄

oyvindio:
In general I'm not negative to having this feature as such. That said I think there is some level of complexity in implementing this, and as I mentioned in my previous review I believe there is some more work needed before this PR is in a state where it could be merged. I am not actively working on this myself.

@mika-hjalmarsson as you opened this PR, is this something you are still interested in and/or working on?

PR #199, opened 2023-01-05
Format with `black -l 120 .`

henrik242:
[black](https://github.com/psf/black) provides pretty sane formatting for python, I think.


mortenlj (approved):
There is so much good chance in this PR that I can't be bothered to point out the handful of places where I would have preferred slight changes in the formatting.

:+1:

PR #200, opened 2023-01-13
Format bin/, tests/ and top level files with black

oyvindio:
I think the auto formatted/formatable code with `black` added in https://github.com/fiaas/fiaas-deploy-daemon/pull/199 is quite nice. I tried to use `black` locally and noticed that `tests/**` and some of the top level files weren't formatted yet, so this PR formats those files as well.
I also added black config in `pyproject.toml` so that the same line length (120) is the default.


mortenlj (approved):
:+1:

henrik242 (approved):
👍

PR #204, opened 2023-03-31
Add Kubernetes 1.24, 1.25 and 1.26 as end-to-end test targets

oyvindio:
- Kubernetes 1.24+ require a more recent kind version, so use the latest release version; v0.18.0. Keep kind v0.11.0 for Kubernetes 1.15 only.
- Upgrade Kubernetes 1.21, 1.22, 1.23 target patch versions to match images built for kind v0.18.0.
- To run tests locally; use kind v0.18.0.
- Remove some references to python 2 and an old setuptools version in README.


j18e (approved):
LGTM

PR #210, opened 2023-10-19
Remove 1.15 version from e2e tests

herodes1991:
Hello. 

We have stopped using the 1.15 version of K8s, so I'm deleting this one. Is there any other version that can be deleted? We are already using 1.24+ K8s versions, so from our side, we can delete 1.21, 1.22, and 1.23 :)

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -65,9 +65,6 @@ blocks:
           # E2E_K8S_VERSION_OPTION in tests/conftest.py; it should point to the most recent version used in CI.
           - env_var: K8S_VERSION
             values:
-              - v1.21.14
-              - v1.22.17
-              - v1.23.17
               - v1.24.12
               - v1.25.8
               - v1.26.3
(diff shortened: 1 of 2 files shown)


portega-adv (approved):
We could probably remove the ones under 1.24, as they are officially deprecated.

oyvindio (dismissed):
Thanks for following up and removing this version from the e2e test. 👍 There is a check for Kubernetes v1.15.x which installs an older version of kind here https://github.com/fiaas/fiaas-deploy-daemon/blob/master/bin/install_e2e_dependencies#L6-L8 , which I think can also be removed.

> Is there any other version that can be deleted? We are already using 1.24+ K8s versions, so from our side, we can delete 1.21, 1.22, and 1.23 :)

Removing Kubernetes versions older than v1.24 from the e2e test is fine for us. @birgirst or @PerGon may want to weigh in on that as well.

herodes1991:
Removed more versions and removed the if :) 
I think we will be able to delete the apiextensions_v1beta1 crd now 😄

oyvindio (approved):

ariadnarouco (approved):

ariadnarouco (approved):

PR #213, opened 2023-11-16
Add certificates to Role with a flag

portega-adv:
This change modifies the Role object in the Helm deployment to include cert-manager.io/v1/Certificate in the allowed Kubernetes objects. 

To keep compatibility, this feature is enabled through a flag (default: false)

### DIFF
--- a/helm/fiaas-deploy-daemon/templates/role.yaml
+++ b/helm/fiaas-deploy-daemon/templates/role.yaml
@@ -54,4 +54,7 @@ rules:
   - list
   - update
   - watch
+{{- if .Values.rbac.role.enableCertificates }}
+{{- include "fiaas-deploy-daemon.RoleCertificateRules" }}
+{{- end }}
 {{- end }}
--- a/helm/fiaas-deploy-daemon/templates/role.yaml
+++ b/helm/fiaas-deploy-daemon/templates/role.yaml
@@ -55,6 +55,6 @@ rules:
   - update
   - watch
 {{- if .Values.rbac.role.enableCertificates }}
-{{- include "fiaas-deploy-daemon.RoleCertificateRules" . }}
+{{ include "fiaas-deploy-daemon.RoleCertificateRules" . }}
 {{- end }}
 {{- end }}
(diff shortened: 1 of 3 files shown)


herodes1991 (dismissed):

helm/fiaas-deploy-daemon/templates/role.yaml
```diff
@@ -54,4 +54,7 @@ rules:
   - list
   - update
   - watch
+{{- if .Values.rbac.role.enableCertificates }}
+{{- include "fiaas-deploy-daemon.RoleCertificateRules" }}
+{{- end }}
```
oyvindio, line 59: I'm getting a helm render error when trying to template the chart (you can test by copying `values.yaml.tpl` -> `values.yaml` locally):
```
$ helm3.9.0 template helm/fiaas-deploy-daemon/ --set rbac.role.enableCertificates=true > 213-enableCertificates-helm-template.yaml
Error: template: fiaas-deploy-daemon/templates/role.yaml:58:4: executing "fiaas-deploy-daemon/templates/role.yaml" at <include>: wrong number of args for include: want 2 got 1

Use --debug flag to render out invalid YAML
```

I guess one option could be to skip the helper and put the additional rules directly in the `if`; it is relatively short

oyvindio (changes requested):

helm/fiaas-deploy-daemon/templates/role.yaml
portega-adv: My bad, but there's a one-character fix for that 😀

oyvindio: With the changes in 46c46f1, the role template renders, but it looks like there is a whitespace issue:
```
$ helm3.9.0 template helm/fiaas-deploy-daemon/ --set rbac.role.enableCertificates=true  --show-only templates/role.yaml
---
# Source: fiaas-deploy-daemon/templates/role.yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: Role

...

  verbs:
  - create
  - delete
  - deletecollection
  - get
  - list
  - update
  - watch- apiGroups:
  - cert-manager.io/v1
  resources:
  - certificates
  verbs:
  - get
  - list
  - watch
```

portega-adv: You're right again, sorry about that, didn't notice. Also fixable, I'll triple check this time!!

oyvindio (approved):
LGTM

PR #214, opened 2023-11-17
Upgrade pytest and related dependencies

oyvindio:
pytest and various related plugins were pinned to versions which supported python 2. Upgrade these dependencies to the most recent versions, since python 2 support is no longer relevant.

- There were several breaking changes between pytest versions 3.x -> 7.x. The changes which affected the tests here were mainly:
  - metafunc.addcall -> metafunc.parametrize, or switch to using parametrize decorator where possible
  - request.funcargvalue -> request.getfixturevalue
  - running tests with `py.test` from the command line doesn't seem to work  (at least locally), because of what appears to be an import/sys.path issue. I've changed the tox commands to use `python -m pytest`, which seems to work fine.

- I've disabled generating html test report because of a bug or interaction issue where the number of tests are reported twice (see 9f42cd9cebaf2aeed8a54aafc3bb19bfda99b383 for details). I think it can be enabled again after upgrading pytest-html, flask and its related dependencies to the most recent versions, but I'd prefer to do that in a separate PR.

- Deprecation warnings when running tests were disabled/hidden a long time ago (I think because there were so many). I've enabled these again, and fixed a few of them. The remaining warnings seem mostly related to Python 3.10 compatibility.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,4 @@
+[pytest]
+markers =
+    integration_test: integration/e2e tests which spin up a kind Kubernetes cluster. Requires docker.  (deselect with '-m "not integration_test"')
+
--- a/setup.cfg
+++ b/setup.cfg
@@ -15,7 +15,3 @@ output=build/reports/coverage.xml
 max-line-length=140
 max-complexity=10
 exclude=build,.git,__pycache__,.eggs,dist,.tox
-
-[tool:pytest]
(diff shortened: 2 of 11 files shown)


herodes1991 (approved):
LGTM. Thanks to take care of this @oyvindio! 🏆

mortenlj (approved):
:trophy:

perj (approved):
:+1:

PR #216, opened 2023-11-27
Add permissions for applications/status

herodes1991:
We detected a problem with the last image and the include-status-in-app flag. We are adding the permissions for the applications/status.


oyvindio (approved):
LGTM

portega-adv (approved):

PR #219, opened 2023-12-01
Document include-status-in-app in operator guide

oyvindio:
Documentation for the feature and config flag introduced in #211 / release v1.2.0 seems to be missing from the operator guide; add some information about the config flag and the behavior when it is enabled in the documentation


perj (approved):
:+1:

herodes1991 (approved):

Issue #220, opened 2024-01-10
Add PodDisruptionBudget support


tg90nor:
Having PDB support in FIAAS should allow us to auto upgrade nodepools and turn on cluster autoscaling without causing unacceptable levels of disruption.

My current plan is to have a default of minAvailable=replicas.min, with an option to explicitly set minAvailable or maxUnavailable, but I'm open to input on this.

Depends on fiaas/k8s#125

herodes1991:
I do not know if this can create problems with Singleton Deployments with only 1 replica (min & max) 😅

miguelbernadi:
My takes :
* Singletons (or apps with a single replica) should not have a PDB by default, as that blocks regular operations in the cluster.
* minAvailable requires knowing the number of replicas, and when the application scales in can have the same issues as a single replica deployment.
* maxUnavailable=1 is the most conservative, ensured to be correct (but not optimal) for any workload with more than 1 replica. I'd go with this for a default.

oyvindio:
Good points in the above comments! 

> * Singletons (or apps with a single replica) should not have a PDB by default, as that blocks regular operations in the cluster.

I definitely agree on this point; if replicas.min or replicas.max is 1 there should probably not be a PDB for the application.

> * minAvailable requires knowing the number of replicas, and when the application scales in can have the same issues as a single replica deployment.

I think I may have originally suggested setting minAvailable=replicas.min in our internal discussions earlier, with the reasoning that it would be intuitive to the user that replicas.min should be available. I see that there may be some issues with that configuration though: As you say, when the application is scaled down to replicas.min, minAvailable would be equal to the actual/desired replicas, and it might not be possible to evict any pods within the PodDisruptionBudget. 

> * maxUnavailable=1 is the most conservative, ensured to be correct (but not optimal) for any workload with more than 1 replica. I'd go with this for a default.

I think maxUnavailable=1 could be a reasonable default (with a recommendation for users to set a higher value for applications that can tolerate it).

Another option I can think of might be to (for applications with >1 replicas) set minAvailable or maxUnavailable to some percentage, e.g. 50%. That would be less conservative, also not optimal, not ensured to be correct (in the sense that it generally avoids service disruptions), but may act as an incentive to set a more accurate value. I'm not saying this is my preferred option, just mentioning it here for completeness.

blockjesper:
[...]

PR #221, opened 2024-01-19
Add PodDisruptionBudget support

tg90nor:
https://github.com/fiaas/fiaas-deploy-daemon/issues/220#issuecomment-1893861140

In this implementation there are no configurable options.
Apps with one replica gets no PDB, and the rest gets a PDB with maxUnavailable=1.

If we see that this causes issues we can tune it and/or make it more configurable. But for now I'm hoping to get some benefits for FIAAS operators without users having to learn anything about PDBs.

### DIFF
--- a/fiaas_deploy_daemon/deployer/kubernetes/adapter.py
+++ b/fiaas_deploy_daemon/deployer/kubernetes/adapter.py
@@ -36,7 +36,8 @@ class K8s(object):
     """Adapt from an AppSpec to the necessary definitions for a kubernetes cluster"""
 
     def __init__(
-        self, config, service_deployer, deployment_deployer, ingress_deployer, autoscaler, service_account_deployer, pod_disruption_budget_deployer
+        self, config, service_deployer, deployment_deployer, ingress_deployer,
+        autoscaler, service_account_deployer, pod_disruption_budget_deployer
     ):
         self._version = config.version
         self._enable_service_account_per_app = config.enable_service_account_per_app
--- a/fiaas_deploy_daemon/deployer/kubernetes/adapter.py
+++ b/fiaas_deploy_daemon/deployer/kubernetes/adapter.py
@@ -66,7 +66,6 @@ class K8s(object):
         self._autoscaler_deployer.delete(app_spec)
         self._service_deployer.delete(app_spec)
         self._deployment_deployer.delete(app_spec)
-        self._pod_disruption_budget_deployer.delete(app_spec)
 
     def _make_labels(self, app_spec: AppSpec):
         labels = {
--- a/fiaas_deploy_daemon/deployer/kubernetes/adapter.py
+++ b/fiaas_deploy_daemon/deployer/kubernetes/adapter.py
@@ -27,6 +27,7 @@ from .deployment import DeploymentDeployer
 from .ingress import IngressDeployer
 from .service import ServiceDeployer
 from .service_account import ServiceAccountDeployer
+from .pod_disruption_budget import PodDisruptionBudgetDeployer
 
 LOG = logging.getLogger(__name__)
 
@@ -35,7 +36,7 @@ class K8s(object):
     """Adapt from an AppSpec to the necessary definitions for a kubernetes cluster"""
 
     def __init__(
-        self, config, service_deployer, deployment_deployer, ingress_deployer, autoscaler, service_account_deployer
+        self, config, service_deployer, deployment_deployer, ingress_deployer, autoscaler, service_account_deployer, pod_disruption_budget_deployer
     ):
         self._version = config.version
         self._enable_service_account_per_app = config.enable_service_account_per_app
@@ -44,6 +45,7 @@ class K8s(object):
         self._ingress_deployer: IngressDeployer = ingress_deployer
         self._autoscaler_deployer: AutoscalerDeployer = autoscaler
         self._service_account_deployer: ServiceAccountDeployer = service_account_deployer
+        self._pod_disruption_budget_deployer: PodDisruptionBudgetDeployer = pod_disruption_budget_deployer
 
     def deploy(self, app_spec: AppSpec):
         if _besteffort_qos_is_required(app_spec):
@@ -56,12 +58,14 @@ class K8s(object):
         self._ingress_deployer.deploy(app_spec, labels)
         self._deployment_deployer.deploy(app_spec, selector, labels, _besteffort_qos_is_required(app_spec))
         self._autoscaler_deployer.deploy(app_spec, labels)
+        self._pod_disruption_budget_deployer.deploy(app_spec, selector, labels)
 
     def delete(self, app_spec: AppSpec):
         self._ingress_deployer.delete(app_spec)
         self._autoscaler_deployer.delete(app_spec)
         self._service_deployer.delete(app_spec)
         self._deployment_deployer.delete(app_spec)
+        self._pod_disruption_budget_deployer.delete(app_spec)
 
     def _make_labels(self, app_spec: AppSpec):
         labels = {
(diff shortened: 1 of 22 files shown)


tests/fiaas_deploy_daemon/specs/v3/test_v3factory.py
```diff
@@ -439,6 +439,7 @@ def test_additional_labels_and_annotations(self, load_app_config_testdata, facto
             service={"service/label": "true"},
             pod={"pod/label": "true", "s": "override"},
             status={"status/label": "true"},
+            pod_disruption_budget={"pod-disruption-budget/label": "true"},
```
perj, line 442: Would it make sense to also add a test in the pytest.mark.parametrize annotation?

perj (commented):
Looks good, just a couple of tiny comments.

fiaas_deploy_daemon/deployer/kubernetes/pod_disruption_budget.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+class PodDisruptionBudgetDeployer(object):
+    def __init__(self, owner_references, extension_hook):
+        self._owner_references = owner_references
+        self._extension_hook = extension_hook
+
+    @retry_on_upsert_conflict
+    def deploy(self, app_spec: AppSpec, selector: dict[str, any], labels: dict[str, any]):
+        if app_spec.autoscaler.min_replicas == 1 or app_spec.autoscaler.max_replicas == 1:
```
perj, line 39: I just want to double check this should be `or` ? The discussions I heard revolved around singletons, which would be `and`. But this might be correct regardless?

perj (approved):
LGTM with a question.

fiaas_deploy_daemon/deployer/kubernetes/pod_disruption_budget.py
tg90nor: We don't want a PDB if there is any circumstance where the app might have a single replica, as that would stall eviction of the pod

fiaas_deploy_daemon/deployer/kubernetes/pod_disruption_budget.py
```diff
@@ -0,0 +1,69 @@
...
+            return
```
oyvindio, line 40: I think if a PodDisruptionBudget for the application already exists (e.g. application was previously deployed with > 1 replicas), it should be deleted here

fiaas_deploy_daemon/deployer/kubernetes/adapter.py
```diff
@@ -56,12 +59,14 @@ def deploy(self, app_spec: AppSpec):
...
         self._deployment_deployer.deploy(app_spec, selector, labels, _besteffort_qos_is_required(app_spec))
         self._autoscaler_deployer.deploy(app_spec, labels)
+        self._pod_disruption_budget_deployer.deploy(app_spec, selector, labels)
 
     def delete(self, app_spec: AppSpec):
         self._ingress_deployer.delete(app_spec)
         self._autoscaler_deployer.delete(app_spec)
         self._service_deployer.delete(app_spec)
         self._deployment_deployer.delete(app_spec)
+        self._pod_disruption_budget_deployer.delete(app_spec)
```
oyvindio, line 69: This is only called on delete watch events (Application resource was deleted). It shouldn't be necessary to add the delete call here since owner references should take care of cleaning up the managed resources. 
(This behavior is left over from before owner references were set on the managed resources, and should probably be cleaned up for other the resources as well, but that's not in scope here )

fiaas_deploy_daemon/deployer/kubernetes/pod_disruption_budget.py
```diff
@@ -0,0 +1,69 @@
...
+
+        custom_annotations = {}
+        custom_labels = labels
+        custom_labels = merge_dicts(app_spec.labels.pod_disruption_budget, custom_labels)
+        custom_annotations = merge_dicts(app_spec.annotations.pod_disruption_budget, custom_annotations)
```
oyvindio, line 45: nitpick: the call to `merge_dicts` here seems unnecessary when `custom_annotations` is always empty

tests/fiaas_deploy_daemon/deployer/kubernetes/test_adapter.py
```diff
@@ -51,6 +52,10 @@ def ingress_deployer(self):
     def autoscaler_deployer(self):
         return mock.create_autospec(AutoscalerDeployer)
 
+    @pytest.fixture(autouse=True)
+    def pod_disruption_budget_deployer(self):
```
oyvindio, line 56: nitpick; It would be nice to add a test verifying that this mock is called on k8s.deploy, similar to the other deployers

oyvindio (commented):
I added a couple of comments in line regarding pdb deletion, other than that I think this looks good.

fiaas_deploy_daemon/deployer/kubernetes/pod_disruption_budget.py
oyvindio: Yes, I think for any case where there can be only one replica (replicas.min=replicas.max=1, with singleton enabled or not; replicas.min=1; replicas.max=1) there shouldn't be a PDB with maxUnavailable=1

oyvindio:
Should it be possible to disable PDBs at the fiaas-deploy-daemon instance level?

tg90nor:
> Should it be possible to disable PDBs at the fiaas-deploy-daemon instance level?

I personally don't think that's necessary, but I'm certainly open to changing my position if there are compelling arguments.

One can always rollback to a previous fdd version if pdbs causes problems, then we can fix those problems.

j18e:
> > Should it be possible to disable PDBs at the fiaas-deploy-daemon instance level?
> 
> I personally don't think that's necessary, but I'm certainly open to changing my position if there are compelling arguments.
> 
> One can always rollback to a previous fdd version if pdbs causes problems, then we can fix those problems.

In the past we've tended to add command line flags when we add new features. We did this when we added support for serviceaccounts. So it wouldn't be out of the ordinary to allow opt-in to the feature.

tg90nor:
> > > Should it be possible to disable PDBs at the fiaas-deploy-daemon instance level?
> > 
> > 
> > I personally don't think that's necessary, but I'm certainly open to changing my position if there are compelling arguments.
> > One can always rollback to a previous fdd version if pdbs causes problems, then we can fix those problems.
> 
> In the past we've tended to add command line flags when we add new features. We did this when we added support for serviceaccounts. So it wouldn't be out of the ordinary to allow opt-in to the feature.

I personally think we have too many flags already, and it's not clear to me that adding it brings any value. With the current FDD release model it is easy to delay introduction of new features by pinning the version.

oyvindio:
> > In the past we've tended to add command line flags when we add new features. We did this when we added support for serviceaccounts. So it wouldn't be out of the ordinary to allow opt-in to the feature.
> 
> I personally think we have too many flags already, and it's not clear to me that adding it brings any value. With the current FDD release model it is easy to delay introduction of new features by pinning the version.

Fair point, part of the reason why most new features were added behind config flags was the previous "rolling release" model. I didn't have a particular use case in mind besides it could perhaps be interesting to not have PDBs in some test environments. We can always add an opt out flag at a later point if we see the need for it.

fiaas_deploy_daemon/specs/v3/factory.py
```diff
@@ -220,6 +220,7 @@ def _labels_annotations_spec(labels_annotations_lookup, overrides):
             "service_account": dict(labels_annotations_lookup["service_account"]),
             "pod": dict(labels_annotations_lookup["pod"]),
             "status": {},
+            "pod_disruption_budget": {},
```
oyvindio, line 223: I missed this on my first review; is there any reason not to support setting labels or annotations on pod disruption budget from the FIAAS config (via labels_annotations_lookup here)? While I can't think of a particular reason it is explicitly needed, it is supported for the other managed resources (except for the (application)status, which is a bit different anyway)

tg90nor: Just an oversight. I added it now.

oyvindio (approved):
Looks good, nice work! 👍

PR #222, opened 2024-01-22
Upgrade k8s to 0.24.2.

perj:
The latest k8s release contains a bug fix for the Watcher class where it could previously hang for up to an hour before noticing a dropped connection. The fix also adds support for bookmark events, which avoids the need of a full quorum read on reconnections in the watcher.


blockjesper (approved):
LGTM

oyvindio (approved):

herodes1991 (approved):

PR #223, opened 2024-01-23
Add PDB rbac rule to helm chart

tg90nor:
I forgot to update the rbac rule in the helm chart when adding PodDisruptionBudget support. This PR fixes that.


blockjesper (approved):
LGTM

Issue #228, opened 2024-01-26
Update Deployments to apps/v1 from extensions/v1beta1


blockjesper:
The extensions/v1beta1 Deployment APIs were removed in Kubernetes 1.16 ([source](https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/)). As all the versions of Kubernetes that we support and test are newer than this, we can safely remove support for managing Deployments via the v1beta1 Deployment API.

PR #230, opened 2024-02-01
Update apiVersion to extensions/v1 in deployment files

blockjesper:
For issue #228

Upgrading apiVersion from extensions/v1beta to extensions/v1.


blockjesper:
I'll try again.

Issue #208, opened 2023-07-03
Fiaas-deploy-daemon creating Role Bindings in addition to Service Account


herodes1991:
We want to improve fiaas-deploy-daemon to create a role-binding in addition to service-account.

We are thinking of adding two configuration flags:
- To enable the creation of the role binding (i.e: --enable-role-binding-creation, by default will be false)
- The list of roles that should be bound for the role binding (i.e: `--list-of-cluster-roles=["A", "B"]` or `--list-of-roles=["A", "B"]`)

I created this issue to start discussing that. I would like to know your opinion 😄

oyvindio:
This feature as it is described sounds reasonable enough to me, assuming the actual roles or clusterroles will be managed elsewhere.

PR #235, opened 2024-02-27
Enable role binding

imorato:
## Description

This Pull Request introduces the functionality to deploy `RoleBindings` in Kubernetes using the `RoleBindingDeployer` from `fiaas_deploy_daemon`. It implements the ability to generate a `RoleBinding` with a `kind` type (`Role` or `ClusterRole`) based on the provided list. This is a PR to fix the issue https://github.com/fiaas/fiaas-deploy-daemon/issues/208

### Changes

1. Added the `RoleBindingDeployer` class in `fiaas_deploy_daemon.deployer.kubernetes.role_binding` to handle the creation and deletion of `RoleBindings` in Kubernetes.

2. Implemented the `_create_role_bindings` method in `RoleBindingDeployer` to create the corresponding `RoleBindings`. The type of `RoleBinding` is determined based on the provided list of roles: if a list of `ClusterRoles` is provided, a `ClusterRole` type `RoleBinding` will be generated, otherwise a `Role` type `RoleBinding` will be created.

### Tests

- Unit tests in `tests/fiaas_deploy_daemon/deployer/kubernetes/test_role_binding.py` to verify the correct functionality of `RoleBindingDeployer` in both scenarios.


bin/run_fdd_against_kind
```diff
@@ -35,6 +35,13 @@ ARGS=(
 '--enable-service-account-per-app'
 '--use-networkingv1-ingress'
 '--use-apiextensionsv1-crd'
+'--enable-role-binding-creation'
+'--list-of-roles' 'RoleA'
+'--list-of-roles' 'RoleB'
+'--list-of-roles' 'RoleC'
+'--list-of-cluster-roles' 'RoleF'
+'--list-of-cluster-roles' 'RoleG'
+
```
herodes1991, line 44: Can we delete those inputs? 
This was created for testing it in local so it's not needed to be there 😄

fiaas_deploy_daemon/config.py
```diff
@@ -143,6 +143,10 @@
 but FIAAS will not overwrite/manage the service account.
 """
 
+ENABLE_ROLE_BINDING_CREATION = """
+Create a role binding for each deployed application, using the application name.
```
herodes1991, line 147:
```suggestion
Create role bindings for the deployed application, one per role or role binding provided, using the application name.
```

fiaas_deploy_daemon/specs/models.py
```diff
@@ -110,6 +110,8 @@ def version(self):
         "pod",
         "status",
         "pod_disruption_budget",
+        "role_binding",
+
```
herodes1991, line 114:
```suggestion
```

tests/fiaas_deploy_daemon/deployer/kubernetes/test_role_binding.py
```diff
@@ -0,0 +1,75 @@
...
+                    call(f"{app_spec.name}-2", app_spec.namespace)
+                ])
+                mock_save.assert_has_calls([
+                    call(),
+                    call()
+                ])
+                owner_references.apply.assert_called()
+
+        _test_create_bindings(["test-role-1", "test-role-2"], "Role")
+        _test_create_bindings(["cluster-role-1", "cluster-role-2"], "ClusterRole")
```
herodes1991, line 42: Can you try to use a parametrize here?
See https://github.com/fiaas/fiaas-deploy-daemon/blob/master/tests/fiaas_deploy_daemon/deployer/kubernetes/test_ready_check.py#L128-L137 as example

herodes1991 (commented):
Small comments

bin/run_fdd_against_kind
imorato: I removed those inputs!!

tests/fiaas_deploy_daemon/deployer/kubernetes/test_role_binding.py
imorato: Nice!!, I've parameterized it now. Take a look and let me know if you need any further changes

perj (commented):
Hi! Sorry that I'm bringing up some discussion that should probably have been on the issue.

I'm slightly curious about the real world use-case for this. Would it be possible to add an example to the documentation?

You'll also need to update `helm/fiaas-deploy-daemon/templates/role.yaml`, I believe.

herodes1991:
> Hi! Sorry that I'm bringing up some discussion that should probably have been on the issue.
>
> I'm slightly curious about the real world use-case for this. Would it be possible to add an example to the documentation?
> 
> You'll also need to update helm/fiaas-deploy-daemon/templates/role.yaml, I believe.

@perj, right now, a field in FIAAS spec creates a service account for the application. This service account does not have permissions by default, so we are working on a part for the SA, creating role bindings to bring some permissions to the FIAAS application. 

I do not know what you mean by "add an example to the documentation": do you want to have the example of the role bindings created with an arbitrary list-of-roles and list-of-cluster-roles provided? 🤔

perj:
I was simply thinking of having an example of when you would want to do this, it doesn't seem very common to me that all FIAAS applications should share the same role binding. But it's not a requirement, I just figured it'd be useful to put it in context. I was thinking of a sentence like "this is useful for when you wish all applications to have read access to system X", or something like that.

fiaas_deploy_daemon/deployer/kubernetes/role_binding.py
```diff
@@ -0,0 +1,61 @@
...
+        self._create_role_bindings(app_spec, self._config.list_of_cluster_roles, "ClusterRole", 1 + len(self._config.list_of_roles))
+
+    def _create_role_bindings(self, app_spec: AppSpec, roles_list, role_kind, counter):
+        LOG.info("Creating RoleBindings for %s", app_spec.name)
+        namespace = app_spec.namespace
+
+        if self._config.enable_service_account_per_app:
+            service_account_name = app_spec.name
+        else:
+            service_account_name = "default"
```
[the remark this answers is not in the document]
herodes1991: Probably you are right, and we can avoid the creation of role bindings if the service_account per app is disabled 🤔

fiaas_deploy_daemon/deployer/kubernetes/role_binding.py
```diff
@@ -0,0 +1,61 @@
...
+
+        for i, role_name in enumerate(roles_list):
```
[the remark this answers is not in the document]
imorato: You're right, it's not being used. I have changed the code to loop directly with the list

fiaas_deploy_daemon/deployer/kubernetes/role_binding.py
```diff
@@ -0,0 +1,61 @@
...
+        try:
+            RoleBinding.delete(f"{app_spec.name}", app_spec.namespace)
+        except NotFound:
+            pass
+
+    def _owned_by_fiaas(self, service_account):
+        return any(
+            ref.apiVersion == "fiaas.schibsted.io/v1" and ref.kind == "Application"
+            for ref in service_account.metadata.ownerReferences
+        )
```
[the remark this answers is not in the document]
imorato: Yes, it was copied from service_account, if it's not needed, I'll remove it :)

herodes1991:
Hey @perj! We've just committed the changes! Thank you for the review!
Can you re-check it? 😄

fiaas_deploy_daemon/deployer/kubernetes/role_binding.py
```diff
@@ -0,0 +1,45 @@
...
+
+        for role_name in roles_list:
+            role_binding_name = f"{app_spec.name}-{counter}"
+            try:
+                role_binding = RoleBinding.get(role_binding_name, namespace)
+            except NotFound:
+                role_binding = RoleBinding()
+
+            role_binding.metadata = ObjectMeta(
+                name=role_binding_name, namespace=namespace
```
oyvindio, line 35: The common labels generated in adapter.py as well as labels and annotations from `app_spec.labels.role_binding` and `app_spec.annotations.role_binding` should be set here (see the other resource deployers)

fiaas_deploy_daemon/deployer/kubernetes/role_binding.py
```diff
@@ -0,0 +1,45 @@
...
+            )
+
+            role_ref = RoleRef(kind=role_kind, apiGroup="rbac.authorization.k8s.io", name=role_name)
+            subject = Subject(kind="ServiceAccount", name=service_account_name, namespace=namespace)
+            role_binding.roleRef = role_ref
+            role_binding.subjects = [subject]
+            self._owner_references.apply(role_binding, app_spec)
+            role_binding.save()
+
+            counter += 1
```
oyvindio, line 45: Two points about the deployer logic here:

1) If a existing role in `list-of-roles` or `list-of-cluster-roles` is removed, the role bindings binding that role to deployed applications service accounts are not removed. I think in that case the (cluster)rolebindings probably should be deleted, to avoid applications retaining privileges they should not have.

2) Because [roleRef is immutable](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#rolebinding-and-clusterrolebinding), the rolebinding naming scheme used here will cause problems when there are changes to the order of the configured roles or clusterroles. Let's say fiaas-deploy-daemon runs with `list-of-roles: [foo, bar]`. Deployed applications have bindings `$app-1` for role `foo` and `$app-2` for role `bar`. If `foo` is removed from the configuration, on the next deploy fiaas-deploy-daemon will try to update rolebinding `$app-1` to bind role `bar`, and the deploy will fail because the API will reject changing the roleRef for that resource. 

Maybe one way to handle these cases may be e.g. to have the rolebinding deployer list all rolebindings and clusterrolebindings with selector `app=$appname,fiaas/deployed_by` (see comment about labels above), compare the desired roles and clusterroles with the existing roles, update the bindings which bind desired roles/clusterroles and delete bindings for roles which are no longer desired. Maybe also do a [check similar to for serviceaccounts](https://github.com/fiaas/fiaas-deploy-daemon/blob/master/fiaas_deploy_daemon/deployer/kubernetes/service_account.py#L78-L82) to avoid accidentally deleting bindings managed by other systems. There are probably other ways to do it as well, this was just my first thought.

helm/fiaas-deploy-daemon/templates/role.yaml
```diff
@@ -57,6 +58,7 @@ rules:
   - pods
   - poddisruptionbudgets
   - resourcequotas
+  - rolebindings
```
oyvindio, line 61: Creating rolebindings have [special RBAC requirements](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#restrictions-on-role-binding-creation-or-update) to avoid privilege escalation. fiaas-deploy-daemon can only bind roles which grant privileges its own serviceaccount already has, unless it is also granted the `bind` verb on the `role` and `clusterrole` resource, which effectively allows privilege escalation. 
Having `bind` on *all* roles and (especially) clusterroles is very powerful since an actor with that privilege can effectively bind e.g. `cluster-admin` to their own user/serviceaccount. 
If the bind role/clusterrole privilege should be added here, I think it would be best to limit it to only a configurable list of named roles and clusterroles (which should usually match the roles and clusterroles set in `list-of-roles` and `list-of-cluster-roles` respectively)

oyvindio (changes requested):
The changes here look mostly good, but there are some special requirements for rolebindings which I think call for some more advanced logic in the deployer, as well as the option to configure additional RBAC privileges for the fiaas-deploy-daemon serviceaccount. See inline comments for details

fiaas_deploy_daemon/deployer/kubernetes/adapter.py
```diff
@@ -60,12 +64,15 @@ def deploy(self, app_spec: AppSpec):
         self._deployment_deployer.deploy(app_spec, selector, labels, _besteffort_qos_is_required(app_spec))
         self._autoscaler_deployer.deploy(app_spec, labels)
         self._pod_disruption_budget_deployer.deploy(app_spec, selector, labels)
+        if self._enable_role_binding_creation is True:
+            self._role_binding_deployer.deploy(app_spec)
```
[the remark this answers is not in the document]
oyvindio: Ah, yes disregard this 👍  I started my review before the latest changes and thought this comment was removed

helm/fiaas-deploy-daemon/templates/role.yaml
herodes1991: > Perhaps this rolebindings permission can also be conditional on that list being non-empty.
I think I will prefer to fail here, than giving such power to fiaas-deploy-daemon pod 😅

fiaas_deploy_daemon/deployer/kubernetes/role_binding.py
herodes1991: Hey @oyvindio, I've just changed the implementation. Can you check? 😄 

I want to discuss first if we are happy with this new implementation before starting to work on the tests 😄 

Right now, I'm getting the list of role_bindings with the label `app: app_name` and then I delete those `owned_by_fiaas` that do not have a role or cluster role provided.

Then, I iterate over the roles and role_bindings and, for any of them, first, I check in the list if exists any role_binding with the expected role or cluster role. I just update it if it exists and it's `owned_by_fiaas`. If it exists and it's not owned_by_fiaas, I log a message.

If it does not exist, I search for an `app-name-{i}` that does not exist and just create it 😄

fiaas_deploy_daemon/specs/v3/defaults.yml
```diff
@@ -93,6 +94,7 @@ labels: # Labels to be set on generated objects
   service: {}
   pod: {}
   pod_disruption_budget: {}
+  role_binding: {}
```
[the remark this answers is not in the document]
herodes1991: Added!

fiaas_deploy_daemon/deployer/kubernetes/role_binding.py
herodes1991: Added!

helm/fiaas-deploy-daemon/templates/role.yaml
herodes1991: Added the roles and cluter_roles to assign to fiaas-deploy-daemon instead of role binding permissions 😄

fiaas_deploy_daemon/deployer/kubernetes/role_binding.py
```diff
@@ -0,0 +1,100 @@
...
+        self._owner_references: OwnerReferences = owner_references
+        self._list_of_roles = config.list_of_roles
+        self._list_of_cluster_roles = config.list_of_cluster_roles
+
+    def deploy(self, app_spec: AppSpec, labels):
+        custom_annotations = {}
+        custom_labels = labels
+        custom_labels = merge_dicts(app_spec.labels.role_binding, custom_labels)
+        custom_annotations = merge_dicts(app_spec.annotations.role_binding, custom_annotations)
+        role_bindings = RoleBinding.find(name=app_spec.name, namespace=app_spec.namespace)
```
perj, line 24: Might want to comment here that `name` refers to the `app` label value. It didn't make sense to me until I looked up the implementation of the find function.

fiaas_deploy_daemon/deployer/kubernetes/role_binding.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def _owned_by_fiaas(self, role_binding):
+        return any(
+            ref.apiVersion == "fiaas.schibsted.io/v1" and ref.kind == "Application"
+            for ref in role_binding.metadata.ownerReferences
+        )
+
+    def _check_if_exists(self, role_binding):
+        return (not (role_binding.roleRef.kind == "Role" and role_binding.roleRef.name in self._list_of_roles)
+                and not (role_binding.roleRef.kind == "ClusterRole" and role_binding.roleRef.name in self._list_of_cluster_roles))
```
perj, line 100: Please add an explanation comment for what this function does, as it's not obvious.
It might make more sense to use a match statement to clarify the cases.

fiaas_deploy_daemon/deployer/kubernetes/role_binding.py
```diff
@@ -0,0 +1,100 @@
...
+                            name=role_binding_name, namespace=namespace, labels=custom_labels, annotations=custom_annotations
+                        )
+
+                        role_ref = RoleRef(kind=role_kind, apiGroup="rbac.authorization.k8s.io", name=role_name)
+                        subject = Subject(kind="ServiceAccount", name=service_account_name, namespace=namespace)
+                        role_binding.roleRef = role_ref
+                        role_binding.subjects = [subject]
+                        self._owner_references.apply(role_binding, app_spec)
+                        role_binding.save()
+                        counter += 1
```
perj, line 76: Although extracting to a method would help, I still think the flow would be much more clear if this part was moved outside of the loop. Should just be to move the break and change the indentation.

fiaas_deploy_daemon/deployer/kubernetes/role_binding.py
```diff
@@ -0,0 +1,100 @@
...
+                if self._owned_by_fiaas(role_binding):
+                    role_binding.metadata = ObjectMeta(
+                        name=role_binding.metadata.name, namespace=namespace, labels=custom_labels, annotations=custom_annotations
+                    )
+                    role_ref = RoleRef(kind=role_kind, apiGroup="rbac.authorization.k8s.io", name=role_name)
+                    subject = Subject(kind="ServiceAccount", name=service_account_name, namespace=namespace)
+                    role_binding.roleRef = role_ref
+                    role_binding.subjects = [subject]
+                    self._owner_references.apply(role_binding, app_spec)
+                    role_binding.save()
```
perj, line 50: There's nine lines here that are identical and need to be kept in sync with below. I think that's enough to warrant extracting it to a separate method.

perj (commented):
Code ooks ok to me. :+1:

I still think you need the permission create rolebinding, even when you have the roles you wish to bind. @oyvindio was explaining what was required in addition to the permisson.

fiaas_deploy_daemon/deployer/kubernetes/role_binding.py
herodes1991: Extracted to a separate method and moved it

fiaas_deploy_daemon/deployer/kubernetes/role_binding.py
herodes1991: Extracted to a separate method and moved it

fiaas_deploy_daemon/deployer/kubernetes/role_binding.py
herodes1991: I added a comment and changed the method name to `check_if_matches`

fiaas_deploy_daemon/deployer/kubernetes/role_binding.py
herodes1991: Added a comment 😄

herodes1991:
> I still think you need the permission create rolebinding, even when you have the roles you wish to bind. @oyvindio was explaining what was required in addition to the permission.

Added again, I totally misunderstood his message 😅

herodes1991:
Working on the tests 😄

oyvindio (changes requested):
Thanks for addressing the previous comments! I did some testing with the current changeset locally and ran into a couple of issues, see the inline comments for details

helm/fiaas-deploy-daemon/templates/custom_rolebinding.yaml
```diff
@@ -0,0 +1,51 @@
+{{- if .Values.rbac.roleBinding.roles -}}
```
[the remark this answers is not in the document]
herodes1991: Fixed the bindings 😄

fiaas_deploy_daemon/deployer/kubernetes/role_binding.py
```diff
@@ -0,0 +1,101 @@
...
+            else:
+                while (True):
+                    role_binding_name = f"{app_spec.name}-{counter}"
+                    try:
+                        role_binding = RoleBinding.get(role_binding_name, namespace)
+                        counter += 1
+                    except NotFound:
+                        role_binding = RoleBinding()
+                        counter += 1
+                        break
```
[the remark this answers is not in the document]
herodes1991: Moved to generateName

fiaas_deploy_daemon/deployer/kubernetes/role_binding.py
```diff
@@ -0,0 +1,101 @@
...
+            if role_binding.roleRef.kind == role_kind and role_binding.roleRef.name == role_name:
+                return role_binding
+            return None
+
+    def _clean_not_needed_role_bindings(self, role_bindings: list[RoleBinding]):
+        for role_binding in role_bindings:
+            if not self._check_if_matches(role_binding):
+                if self._owned_by_fiaas(role_binding):
+                    RoleBinding.delete(role_binding.metadata.name, role_binding.metadata.namespace)
+                    role_bindings.remove(role_binding)
```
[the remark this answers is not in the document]
herodes1991: iterating a copy list and removing only from the original one

fiaas_deploy_daemon/deployer/kubernetes/role_binding.py
```diff
@@ -0,0 +1,101 @@
...
+    def _owned_by_fiaas(self, role_binding):
+        return any(
+            ref.apiVersion == "fiaas.schibsted.io/v1" and ref.kind == "Application"
+            for ref in role_binding.metadata.ownerReferences
+        )
+
+    # Check if matches the role_binding with any role or clusterRole in the list_of_roles or list_of_cluster_roles
+    def _check_if_matches(self, role_binding):
+        return (not (role_binding.roleRef.kind == "Role" and role_binding.roleRef.name in self._list_of_roles)
+                and not (role_binding.roleRef.kind == "ClusterRole" and role_binding.roleRef.name in self._list_of_cluster_roles))
```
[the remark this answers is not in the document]
herodes1991: fixed the if, I just swapped the logic by error 🤦

fiaas_deploy_daemon/deployer/kubernetes/role_binding.py
```diff
@@ -0,0 +1,101 @@
...
+        for role_binding in role_bindings:
+            if role_binding.roleRef.kind == role_kind and role_binding.roleRef.name == role_name:
+                return role_binding
+            return None
+
+    def _clean_not_needed_role_bindings(self, role_bindings: list[RoleBinding]):
+        for role_binding in role_bindings:
+            if not self._check_if_matches(role_binding):
+                if self._owned_by_fiaas(role_binding):
+                    RoleBinding.delete(role_binding.metadata.name, role_binding.metadata.namespace)
```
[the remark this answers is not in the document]
herodes1991: Added!

tests/fiaas_deploy_daemon/deployer/kubernetes/test_role_binding.py
```diff
@@ -0,0 +1,121 @@
...
+             patch.object(RoleBinding, 'save') as mock_try_to_save:
+
+            deployer._update_or_create_role_bindings(app_spec, roles_list, role_kind, 1, {}, {}, [])
+            deployer._update_or_create_role_bindings(app_spec, cluster_roles_list, cluster_role_kind, 1, {}, {}, [])
+
+            mock_get_try_to_create.assert_not_called()
+            mock_try_to_save.assert_not_called()
+            owner_references.apply.assert_not_called()
+
+    def test_deploy_full_flow(self, deployer, owner_references, app_spec):
```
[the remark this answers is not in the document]
herodes1991: Yes, it was some errors in the implementation, sorry so much for that 😔 

Now it should be fixed. Added also some tests to ensure the correct work of clean and deploy methods

herodes1991:
Did some tests locally and seems to work, but a second review is more than welcome 😄

oyvindio:
> Did some tests locally and seems to work, but a second review is more than welcome 😄

I will try to take another look in the next couple of days

helm/fiaas-deploy-daemon/templates/custom_rolebinding.yaml
```diff
@@ -0,0 +1,95 @@
...
+{{- end }}
+rules:
+  - apiGroups:
+    - rbac.authorization.k8s.io
+    resources:
+    - clusterroles
+    verbs:
+    - bind
+    resourceNames:
+    {{- range $clusterRole := .Values.rbac.roleBinding.cluserRoles }}
```
oyvindio, line 71:
```suggestion
    {{- range $clusterRole := .Values.rbac.roleBinding.clusterRoles }}
```
This looks like a typo?

oyvindio (approved):
Thanks for the latest updates; the implementation looks good to me now. I also ran some tests locally and it seems to work like it should.

Some minor nitpicks regarding the helm values/documentation (see inline comments), other than that this seems ready to me. 👍

helm/fiaas-deploy-daemon/templates/custom_rolebinding.yaml
herodes1991: Yes, was a typo, probably it was wrong written in my values 🙈

docs/operator_guide.md
```diff
@@ -275,6 +275,17 @@ specified as `extensions.secrets.foosecrets.annotations` will be added to the po
...
+### RoleBinding and ClusterRole Configuration (--list-of-cluster-roles --list-of-roles)
+
+Enables the creation of a `RoleBinding`. If provided a list-of-cluster-roles, it will generate the `RoleBinding` with the kind `ClusterRole`. If provided a list-of-roles, it will generate the `RoleBinding` with the kind `Role`.
+This is useful for when you wish all applications to have read access to specific resources.
+
+##### List of ClusterRoles and Roles for Role Binding
+
+- `--list-of-cluster-roles=["A", "B"]`: This flag allows the creation of `RoleBinding` objects with `ClusterRole` kind for the specified cluster-level roles.
+
+- `--list-of-roles=["A", "B"]`: This flag allows the creation of `RoleBinding` objects with `Role` kind for the specified namespace-level roles.
```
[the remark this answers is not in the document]
herodes1991: True, working on adding the advice here and working on the empty list in the values.yml.tpl file 😄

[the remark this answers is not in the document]
herodes1991: Done! Thank you so much for your review!

oyvindio (dismissed):

perj (approved):

portega-adv (approved):

PR #238, opened 2024-04-09
Handle exceptions from lifecycle.failed

oyvindio:
An edge case with a combination of a FIAAS config with a lot (thousands) of ingress paths and fiaas-deploy-daemon configured with several ingress-suffix hostnames can generate an ingress that is large enough to exceed the Kubernetes API maximum request payload limit of 3MB. This also causes the ApplicationStatus resource update (via `lifecycle`) to exceed the same limit, since it holds the log entries generated while deploying the application, and the AppSpec object is logged in full at least twice.
If this happens, the deployer and/or scheduler threads exit because exceptions caused by the failed API requests flow unhandled out of these. The liveness probe for fiaas-deploy-daemon then fails, causing the pod to be restarted.

In theory similar exceptions caused by other failure modes can also cause either of these threads to exit, so to increase robustness; log and drop exceptions from failed Kubernetes API requests while updating ApplicationStatus in the exception handler in the deployer thread, and where the scheduler thread calls ReadyChecks. Application deployments that trigger exceptions here will still fail in a half-deployed RUNNING state, but at least fiaas-deploy-daemon will not restart each time it attempts to deploy an application with a configuration like described above.

### DIFF
--- a/fiaas_deploy_daemon/deployer/deploy.py
+++ b/fiaas_deploy_daemon/deployer/deploy.py
@@ -19,6 +19,9 @@
 import logging
 from queue import Queue
 
+from k8s.client import K8sClientException
+from requests.exceptions import RetryError
+
 from fiaas_deploy_daemon.config import Configuration
 from fiaas_deploy_daemon.deployer.kubernetes.adapter import K8s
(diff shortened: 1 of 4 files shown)


blockjesper (approved):
LGTM

herodes1991 (approved):
Good catch!

Issue #104, opened 2020-07-03
Option to disable service links environment variables


xavileon:
k8s injects a bunch of service links variables to be able to reach other services within a namespace via env vars. This feature is deprecated since a lot time ago. However, those environment variables are being injected anyway which makes some services to fail (e.g specific example being nodejs apps).

Disabling this environment variables injection is already possible in k8s 1.13 via the `enableServiceLinks` flag in the [PodSpec](https://v1-15.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#podspec-v1-core).

It would be helpful to get this flag configurable (possibly default to false? given it shouldn't be used as it's a deprecated feature).

tg90nor:
I'm ok with implementing this as a configurable flag. Just keep in mind fiaas-deploy-daemon is in use in clusters with kubernetes 1.12, so it needs to work for those. I also thing it should default to true, since that is what kubernetes is doing.

xavileon:
great! yeah, I think FIAAS is currently supporting k8s versions starting from 1.9 according to [tests](https://github.com/fiaas/fiaas-deploy-daemon/blob/master/tests/fiaas_deploy_daemon/conftest.py#L192), so the implementation should handle different k8s versions indeed.

mortenlj:
There is probably some value in checking if we still need to support the oldest versions.

oyvindio:
It is generally not a good idea to use the service host/port environment variables, so I think it makes a lot of sense to make it possible to configure `enableServiceLinks` at the fiaas-deploy-daemon instance level.

blockjesper:
Done.

PR #240, opened 2024-04-15
Enable service links

blockjesper:
Add support to disable service links. #104

### DIFF
--- a/fiaas_deploy_daemon/config.py
+++ b/fiaas_deploy_daemon/config.py
@@ -328,6 +328,15 @@ class Configuration(Namespace):
             default="1",
             type=_int_or_unicode
         )
+        parser.add_argument(
+            "--disable-service-links",
+            dest="enable_service_links",
+            help=(
+                "Disable service links in the podspec. Recommended "
+                " for new installations. Existing installations may need to skip this."
+                ),
+            action="store_false"
+        )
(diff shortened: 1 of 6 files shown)


fiaas_deploy_daemon/config.py
```diff
@@ -328,6 +328,15 @@ def _parse_args(self, args):
             default="1",
             type=_int_or_unicode
         )
+        parser.add_argument(
+            "--disable-service-links",
+            dest="enable_service_links",
```
herodes1991, line 333: For me, this is confusing 😅.
If the flag is --disable-service-links, the config should be the same 😄

blockjesper: I agree, but I think this is the reasonable way of doing it due to some unfortunate circumstances.

1) The kubernetes API takes the argument as enableServiceLinks, with default for now being true.
2) The library used for parsing command line arguments doesn't take an argument for boolean arguments, so it's not possible to send in false together with it. My initial attempt was with the argument as "enable-service-links", but hit this problem. 

You do definitely have a point that it's confusing however.

henrik242: Maybe just add a comment explaining this?

blockjesper: @henrik242 I've added a comment about the inverted logic.

PR #254, opened 2024-10-04
Move project definition including dependencies to pyproject.toml

oyvindio:
The primary motivation for this is to get the Github dependency graph to resolve properly so that dependabot security alerts can work in this repository. Apparently the system that resolves the dependency graph [can be unreliable when dependencies are defined in setup.py][1], and [using variables when specifying dependencies may also cause issues][2].

pyproject.toml / PEP621 seems like the(?) new standard of configuring python projects and associated tooling, so I think this is a step in the right direction in any case.

I think most things should work as before, but do let me know if something doesn't work properly in your environment. Some minor differences:
- the `[dev]` dependency target doesn't include the dependencies from the `[ci]` target. To work around that, it is necessary to use `[dev,ci]` to achieve the same thing as `[dev]` previously. requirements.txt is updated to reflect this, so `pip install -r requirements.txt` should still work as before.
- the fiaas-deploy-daemon package is no longer marked as zip-safe. The zip-safe flag is [deprecated][3], and it was only added to work around an old bug in setuptools.

[1]: https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-the-dependency-graph#supported-package-ecosystems
[2]: https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/troubleshooting-the-dependency-graph#does-the-dependency-graph-detect-dependencies-specified-using-variables
[3]: https://setuptools.pypa.io/en/latest/deprecated/zip_safe.html


nerhus (approved):

perj (approved):
:+1:

blockjesper (approved):

PR #258, opened 2024-10-10
Remove bin/ingress_check.py and dnspython dependency

oyvindio:
I don't think `bin/ingress_check.py` is relevant any more, and it is the only thing I can find which uses `dnspython` (I think dnspython was also used in the kafka consumer, but that has been removed for a long time). Remove the script and the dependency on dnspython.

Also remove dependency on `ipaddress`; this has been in the standard library since python 3.3


blockjesper (approved):
LGTM

PR #260, opened 2024-10-18
Add support in PDBs to be able to set unhealthyPodEvictionPolicy.

blockjesper:
This PR introduces support for the unhealthyPodEvictionPolicy to provide more control over pod eviction behavior managed by PodDisruptionBudgets (PDBs). The policy can now be configured per FIAAS deployment daemon instance, allowing administrators to define how unhealthy pods are handled during voluntary disruptions like node maintenance.

**Default setting**: IfHealthyBudget, ensuring evictions only occur if they respect the PDB's disruption budget.
**Optional setting**: AlwaysAllow, can be useful for dev-environments where deployments may be broken or when the applications can tolerate the disruptions.

Documentation: https://kubernetes.io/docs/tasks/run-application/configure-pdb/#unhealthy-pod-eviction-policy

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ limitations under the License.
 
 [![Build status](https://fiaas-svc.semaphoreci.com/badges/fiaas-deploy-daemon.svg?style=shields)](https://fiaas-svc.semaphoreci.com/projects/fiaas-deploy-daemon)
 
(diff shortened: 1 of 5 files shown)


perj (approved):
LGTM

oyvindio (approved):

Issue #261, opened 2024-11-25
Can't use Kubernetes delete cascade orphan


dacofr:
Hi :wave: 

With the current implementation of Fiaas, we cannot use :
`kubectl delete fa bla-bla --cascade=orphan` 

Some objects like PDB, RoleBindings, SA are well orphaned after deletion (owner reference is removed and objects still remain clusters), but that's not the case for Deployment, Service, Ingress because there is an explicit deletion that is done here : 

```
    def delete(self, app_spec: AppSpec):
        self._ingress_deployer.delete(app_spec)
        self._autoscaler_deployer.delete(app_spec)
        self._service_deployer.delete(app_spec)
        self._deployment_deployer.delete(app_spec)
```
https://github.com/fiaas/fiaas-deploy-daemon/blob/master/fiaas_deploy_daemon/deployer/kubernetes/adapter.py#L69

This made sense in the past, but now that each objects has an owner reference, we can let Kubernetes do the garbage collection for us.

I will send a PR to remove this explicit deletion, WDYT ?

Thanks

perj:
Hi @dacofr!

I think that makes a lot of sense to support all three deletion modes.

If I'm reading [the docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/#ownership-and-finalizers) (also [this one](https://kubernetes.io/docs/tasks/administer-cluster/use-cascading-deletion/)) correctly the controller (fiaas-deploy-daemon) should still delete the children in foreground mode.
Thus the code should check the finalizers of the application and still run the above code if it's set to foreground, but not run if it's unset. In orphan mode, it still needs to update the children to remove the ownerReference, I belive, I don't see anything in the docs about that being automatic.

When using foreground of orphan mode, it seems the controller is [expected](https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/) to update the Application object and remove those finalizers. This part also need to be implemented for it to work fully. But I think if the children still has the ownerReference when that happens, they'll still be automatically deleted. Do test that part though.

dacofr:
Hello

@perj and @oyvindio, thanks both for your feedback :+1: 
Yes during my tests, simply commenting out the four lines of explicit deletion was enough to have all three modes of deletion to work as expected. 

Didn't find this existing PR. 
But @oyvindio, If you want, I can start a new fresh PR and do the same kind of stuff.

dacofr:
[...]

PR #262, opened 2024-11-27
Remove Delete event

dacofr:
This PR intend to solve #261 where due to explicit deletion of owned resources (Deployment, Ingress, HPA, Service) we can't use delete cascade orphan mechanism provided by Kubernetes. 

This PR do exactly same stuff as #101:

> Now that we are setting ownerReferences on the created objects
kubernetes will take care of deleting them for us when the application
is deleted, so we can remove all the code related to that.

> The delete methods for service and autoscaler are still needed
for the case where the config changes such that they are no longer
required, but now they can be private/internal methods.

### DIFF
--- a/fiaas_deploy_daemon/crd/watcher.py
+++ b/fiaas_deploy_daemon/crd/watcher.py
@@ -96,9 +96,7 @@ class CrdWatcher(DaemonThread):
         # removed. If the application resource is in the process of being deleted, deploy should be skipped otherwise
         # it may interfere with the deletion process done by the garbage collector.
         deletion_timestamp = application.metadata.deletionTimestamp
-        if isinstance(deletion_timestamp, datetime.datetime) and deletion_timestamp <= datetime.datetime.now(
-            tz=datetime.timezone.utc
-        ):
+        if isinstance(deletion_timestamp, datetime.datetime):
             LOG.warning(
                 "Skipping update watch event for app %s; it was marked for deletion at %s",
                 application.spec.application,


fiaas_deploy_daemon/crd/watcher.py
```diff
@@ -72,8 +72,6 @@ def _watch(self, namespace):
     def _handle_watch_event(self, event: WatchEvent):
         if event.type in (WatchEvent.ADDED, WatchEvent.MODIFIED):
             self._deploy(event.object)
-        elif event.type == WatchEvent.DELETED:
-            self._delete(event.object)
```
perj, line 76: This doesn't seem right, the Kubernetes API will still generate these events and we shouldn't be raising any exception when receiving them.

Now that i check #101 I see it keeps the if check before raising.

tests/fiaas_deploy_daemon/crd/test_crd_watcher.py
```diff
@@ -156,7 +151,6 @@ def test_is_able_to_watch_custom_resource_definition(self, crd_watcher, deploy_q
             (ADD_EVENT, "UPDATE", {"deployment": {"fiaas/source-repository": "xyz"}}, "xyz"),
             (MODIFIED_EVENT, "UPDATE", None, None),
             (MODIFIED_EVENT, "UPDATE", {"deployment": {"fiaas/source-repository": "xyz"}}, "xyz"),
-            (DELETED_EVENT, "DELETE", None, None),
```
perj, line 159: I feel that the test could be reworked and kept, to catch us not raising an exception. But it's a fair rewrite, admittedly. Perhaps make it a separate test.

fiaas_deploy_daemon/crd/watcher.py
dacofr: Good catch, you're right. Updating PR ...

tests/fiaas_deploy_daemon/crd/test_crd_watcher.py
dacofr: Regarding your previous comment, make sense to keep this test to be sure we are not raising an exception :+1:

dacofr: Created a dedicated test for delete event.
Let me know, if you think it's not enough.

tests/fiaas_deploy_daemon/crd/test_crd_watcher.py
```diff
@@ -229,6 +224,11 @@ def test_deploy(
             assert deployer_event == DeployerEvent(deployer_event_type, app_spec, None)
         assert deploy_queue.empty()
 
+    def test_delete(self, deploy_queue, watcher):
+        watcher.watch.return_value = [WatchEvent(DELETED_EVENT, FiaasApplication)]
+
```
perj, line 229: You'll have to call `crd_watcher._watch(None)` here, other than that it looks good.

perj (approved):
:+1:

oyvindio:
Thanks for the PR! I think the changes here in itself look good. However since this behavior isn't that well covered in e2e tests, I did some tests locally and I noticed an interesting edge case:

When a application resource is deleted with the foreground or orphan modes, the resource is first updated with `metadata.deletionTimestamp`, and the `foregroundDeletion` or `orphan` finalizer is added depending on the mode. The update which adds `deletionTimestamp` creates a update watch event, which fiaas-deploy-daemon will act on. If the application's most recent deploy completed successfully, nothing will happen.
 
However if the most recent deploy for the application didn't complete (applicationstatus with the matching `fiaas/deployment`_id label doesn't have `result=SUCCESS`), fiaas-deploy-daemon will redeploy the application at the relative same time as the garbage collector cleans up the dependent resources. Depending on the order of the operations (which we do not control), resources may be re-created after they have been deleted by the garbage collector (in case of foreground mode), or updated after the owner reference on the resources has been removed (in case of orphan mode).
When fiaas-deploy-daemon then re-creates or updates the resources, owner references will be set again. Since the application resource is deleted or soon will be deleted, the owner references on the resources that are created/updated will be broken, so the garbage collector will delete the resources. This means the orphan deletion mode only works properly if the application was deployed completely on the most recent deploy.

One way to avoid this may be to check `metadata.deletionTimestamp` before deploying and skip the event if deletionTimestamp is set and is in the past. I ended up doing that out of curiosity just to see if it would work (it seems like it does), and pushed it in in a spin-off branch here https://github.com/fiaas/fiaas-deploy-daemon/compare/delete-skip-update-with-deletionTimestamp. I think this edge case should be solved, so either if you can think of an an alternate solution that can be an option, or if you want to cherry-pick 1bcc43fe50fb2021bbd3e05d42d57a277c89273f into this branch that should work too.

oyvindio:
> Just one question that comes to my mind, why checking that `deletionTimestamp` is in the past ? Checking not null should be enough 🤔

I don't have a particular reason for checking `metadata.deletionTimestamp` is in the past other than it "seemed correct". I guess in theory the check could fail to catch some cases if clocks on the control plane and the node that fiaas-deploy-daemon runs on aren't in sync. I'm open to changing the check to only checking if `metadata.deletionTimestamp` is a `datetime.datetime` (this was actually the first thing I tried) or even `is not None`.
 
From the [documentation](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/#objectmeta-v1-meta) it seems like the `deletionTimestamp` communicates that the resource will be deleted *before or at* the timestamp it is set to and the field can't be unset (i.e can't abort a delete which has already been triggered), so it could make sense to just check if the field has value.

oyvindio (dismissed):

dacofr:
Hello,

I removed the check about `deletionTimestamp`, I think that since the field is set independently in past/future is enough to skip the deploy process.

It's fine for me, feel free to merge :tada: 
Thanks for your help and quick review process of this PR :heart:

oyvindio (approved):

dacofr:
Last question, what is you release cycle ? Do you think it's possible to release a new version with current PR ?
Thanks

oyvindio:
> Last question, what is you release cycle ? Do you think it's possible to release a new version with current PR ?

Releases are usually just done as needed, I've released [v1.8.0](https://github.com/fiaas/fiaas-deploy-daemon/releases/tag/v1.8.0) with the changes from this PR now.

PR #263, opened 2024-12-20
Add a dns-search-domains config option.

perj:
This sets the podSpec.dnsConfig.searches. The main use-case I have in mind is to search services in different namespaces, e.g. if you have a hierarchical namespace setup it can be used to search the parent namespace automatically.

Internal ticket: CDX-24.

### DIFF
--- a/fiaas_deploy_daemon/config.py
+++ b/fiaas_deploy_daemon/config.py
@@ -356,7 +356,7 @@ class Configuration(Namespace):
             default=[],
             action="append",
             type=str,
-            dest="search_domains",
+            dest="dns_search_domains",
         )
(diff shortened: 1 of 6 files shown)


fiaas_deploy_daemon/config.py
```diff
@@ -345,8 +347,16 @@ def _parse_args(self, args):
...
+            ),
+            action="store_false",
+        )
+        parser.add_argument(
+            "--dns-search-domains",
+            help="list of extra dns search domains",
+            default=[],
+            action="append",
+            type=str,
+            dest="search_domains",
```
oyvindio, line 359: nitpick: maybe consider keeping the same name for the config attribute and the flag to make it easier to find in the code later

oyvindio (dismissed):
Just a minor nitpick, otherwise this looks good 👍

oyvindio (approved):

PR #269, opened 2025-07-02
Add node selector, affinity and toleration handling to the Helm chart

dmegyesi:
It's common to deploy non-user workloads such as the FIAAS daemon on centralized management nodes. By adding fields for node selectors and affinities (and tolerations), we can start using advanced scheduling logic such as required by the Cluster Autoscaler or Karpenter.

### DIFF
--- a/helm/fiaas-deploy-daemon/templates/deployment.yaml
+++ b/helm/fiaas-deploy-daemon/templates/deployment.yaml
@@ -83,18 +83,14 @@ spec:
             name: fiaas-deploy-daemon-config
             readOnly: true
       {{- if .Values.deployment.nodeSelector }}
-      nodeSelector:
-{{ toYaml .Values.deployment.nodeSelector | indent 8 }}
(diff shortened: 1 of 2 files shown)


portega-adv (dismissed):

helm/fiaas-deploy-daemon/templates/deployment.yaml
```diff
@@ -82,3 +82,19 @@ spec:
           - mountPath: /var/run/config/fiaas/
             name: fiaas-deploy-daemon-config
             readOnly: true
+      {{- if .Values.deployment.nodeSelector }}
+      nodeSelector:
+{{ toYaml .Values.deployment.nodeSelector | indent 8 }}
```
perj: This is a bit nitpicky but no other parts of the file move these all the way to the left. 
Would you mind changing them to look like this:
```
      nodeSelector: {{ toYaml .Values.deployment.nodeSelector | nindent 8 }}
```
?

perj (commented):
Looks good otherwise :+1:

perj (approved):

PR #271, opened 2025-09-29
Statefulsets

NiclasLindqvist:
## Summary

- Introduces FIAAS spec v4 with a new statefulset configuration block, defaults, and transformer/validation so v1–v3 configs continue to upgrade seamlessly.
- Adds end-to-end runtime support for stateful workloads: a StatefulSetDeployer, PVC template generation, headless service handling, autoscaler gating, and integration into the existing Kubernetes adapter.
- Updates documentation, sample manifests, and web defaults to describe v4/statefulset usage and provides a “What’s new” guide for adopters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ spec:
   application: example
   image: nginx:1.13.0
   config:
-    version: 3
+    version: 4
     ingress:
       - host: example.com
     metrics:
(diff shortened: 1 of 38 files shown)


perj:
That's very kind of you Niclas, statefulsets are requested with some frequency. I can tell you right now that this isn't really how I imagined them getting implemented, but it's definitely an option.

I don't really see any reason to bump the major spec version for this change though, there's no backwards incompatibility as far as I can see.

I do also think that we need to carefully consider the configuration keys, which might take some iterations. I would rather talk about `storage` then `statefulset` personally. But there's probably multiple other opinions on the best abstractions. We can involve you in those discussions if you wish

NiclasLindqvist:
You’re probably right, the v4 was excessive since its opt-in. I ended up there as the statefulset option didn’t allow for hpa, but yes, it’s not breaking any existing functionality :) 

And if you have other ways planned for it, we can just close this. Interested to see what it will look like in the end.

